PR #3, opened 2018-11-13
Update build documentation and scripts

asauber:
Update build documentation to reference the Linode environment and toolchain.

Replace existing build toolchain with a single Makefile and Dockerfile.


hack/make.py
```diff
@@ -25,9 +25,8 @@ def check_antipackage():
         exit(1)
 check_antipackage()
 
-# ref: https://github.com/ellisonbg/antipackage
 import antipackage
-from github.appscode.libbuild import libbuild, pydotenv
+from github.asauber.libbuild import libbuild, pydotenv
```
sarahelizgray: Should we move https://github.com/asauber/libbuild into our github org since it's a dependency and forked off of apps code?

asauber: I'm in the process of stripping this entire build/deploy toolchain from this repo.

PR #4, opened 2018-11-15
Rewrite main

asauber:
* Strip out the nested Cobra-command based entrypoint
* Use the latest cloudprovider style entrypoint
* Upgrade dependencies
* Upgrade base Kubernetes version to 1.11.3
* Switch from glide to dep!


Gopkg.toml
```diff
@@ -0,0 +1,63 @@
+# Gopkg.toml example
+#
+# Refer to https://golang.github.io/dep/docs/Gopkg.toml.html
```
sarahelizgray: Should we preserve this example code? Is that helpful?

Gopkg.toml
```diff
@@ -0,0 +1,63 @@
...
+# [[override]]
+#   name = "github.com/x/y"
+#   version = "2.4.0"
+#
+# [prune]
+#   non-go = false
+#   go-tests = true
+#   unused-packages = true
+
+required = ["github.com/json-iterator/go"]
```
sarahelizgray: GO dependency n00b question: Do we need this required statement if we then have the same repo specified as a contraint at line 42?

Gopkg.toml
asauber: Not particularly helpful. I'll remove it.

Gopkg.toml
asauber: Don't blame you on that. There are over 15 go dependency management systems ([https://github.com/golang/go/wiki/PackageManagementTools](https://github.com/golang/go/wiki/PackageManagementTools)) This one, `dep`, requires you to specify a package in the "required" array if you would like to impose a constraint on a package that is not directly (but indirectly) required by your application. In this case, the constraint solver was picking a bad version for this package (couldn't compile a dependency code branch). So I had to manually override.

PR #12, opened 2018-12-04
add LINODE_NB_PREFIX environment variable to prefix NodeBalancer labels 

displague:
Adds a `LINODE_NB_PREFIX` variable which, if present, will prefix all created Linode NodeBalancers with that prefix before the autogenerated unique ID.

For example, to create NodeBalancers named "k8s-12abe93d...." (32 characters will be used)

```yaml
          - name: LINODE_NB_PREFIX
            value: k8s-
```

The prefix is restricted to 12 characters to maintain some value from the UUID.

---

**For Compatibility** with earlier versions, set the `LINODE_NB_PREFIX` to `a`.

### DIFF
--- a/cloud/linode/cloud.go
+++ b/cloud/linode/cloud.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"regexp"
 
 	"github.com/linode/linodego"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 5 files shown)


phillc:
Is this still relevant today? It has aged quite a bit, and I don't have context on its need.

displague:
@phillc You'll notice that the CCM generated loadbalancers still create NodeBalancers with a fixed prefix, at the time this PR was created that prefix was "a" and the NodeBalancer name was "a{hash}".

This PR was created to make the "a" prefix configurable.

It appears that the prefix has changed to "ccm-": https://github.com/linode/linode-cloud-controller-manager/blob/62b517ea48fcd510a2575c59ea342e79a7ffdc26/cloud/linode/loadbalancers.go#L458

I'll close this as I don't have the same need and I'm not prepared to rebase this any time soon.

PR #21, opened 2019-04-03
E2E test for simple LoadBalancer Service

sanjid133:
### General:

* [ ] Have you removed all sensitive information, including but not limited to access keys and passwords?

### Pull Request Guidelines:

1. [ ] Have you added tests? 


faem:
### Command to run test:
```
ginkgo --v -- <custom-flags>
```
### Available Flags:
```
--image registry/repository:tag
    image of the linode-csi
--api-token API_TOKEN
    linode api token
--delete
    delete cluster after test
--use-existing
    use existing cluster
--kubeconfig <kube config file path>
    to use existing cluster provide kubeconfig file
```

e2e/test/framework/framework.go
```diff
@@ -0,0 +1,65 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+var (
+	Image    = "linode/linode-cloud-controller-manager:latest"
+	ApiToken = ""
+
+	KubeConfigFile         = ""
+	testServerResourceName = "e2e-test-server-" + rand.Characters(5)
```
asauber: I think we should change this name.

If it will be used for resource types other than services, then the prefix should be "e2e-test-".

If it will only be used for services, then it should be defined in service.go, and the prefix should be "e2e-test-service-"

asauber:
Looks great

PR #33, opened 2019-06-11
Use default linodego http client rather than creating a custom one

jfrederickson:
The oauth2.Transport we were previously creating here was a valid
http.RoundTripper, but was not an http.Transport. Resty (which
linodego also uses) expects the transport to be an http.Transport for
certain features.

Requires https://github.com/linode/linodego/pull/80


asauber:
needs rebase, linodego updated, tests should pass now

PR #39, opened 2019-09-13
Port annotation

phillc:
New port annotation.

Fixes issues with the former protocol annotation where specifying https would not allow for an http port to be opened, as the node balancer creation would fail.


asauber:
Hi @phillc. If this is a duplicate of #46 can we close this one?

PR #40, opened 2019-10-03
Bump linodego to v0.11

adammohammed:
This PR bumps the linodego version to prevent go panics due to bad responses from the LinodeAPI. Now an error is properly raised instead of segfaulting when the API returns a content-type other than "application/json".


phillc:
IMO, either update linodego in the e2e project as well, or merge in #41

adammohammed:
Closing since #41 also bumped version for e2e

PR #43, opened 2019-10-03
go mod tidy

phillc:


jnschaeffer:
Do we want to update this with another `go mod tidy` call and then merge it?

PR #44, opened 2019-10-07
Go.sum had 3 versions of linodego, remove old versions

adammohammed:
This is just a tidying step for the go.sum and go.mod files. The current version of linodego is v0.12, should we bump to that?


adammohammed:
Woops, there is already a PR for this

PR #46, opened 2019-10-15
Port annotation

phillc:


asauber:
Please resolve conflicts with master.

cloud/linode/loadbalancers.go
```diff
@@ -339,31 +345,24 @@ func (l *loadbalancers) buildNodeBalancerConfig(service *v1.Service, port int) (
...
 
 	return config, nil
 }
 
-func (l *loadbalancers) processHTTPS(service *v1.Service, nbConfig *linodego.NodeBalancerConfig, port int) error {
-	if err := l.retrieveKubeClient(); err != nil {
-		return err
-	}
-	tlsAnnotations, err := getTLSAnnotations(service)
+func (l *loadbalancers) addSSLCert(service *v1.Service, nbConfig *linodego.NodeBalancerConfig, portConfig *portConfig) error {
```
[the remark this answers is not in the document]
jnschaeffer: Update: Talked about this offline, we can keep it as-is.

cloud/linode/loadbalancers.go
```diff
@@ -485,26 +484,24 @@ func getNodeInternalIP(node *v1.Node) string {
...
-func getTLSCertInfo(kubeClient kubernetes.Interface, tlsAnnotations []*tlsAnnotation, namespace string, port int) (string, string, error) {
-	for _, tlsAnnotation := range tlsAnnotations {
-		if tlsAnnotation.Port == port {
-			secret, err := kubeClient.CoreV1().Secrets(namespace).Get(tlsAnnotation.TLSSecretName, metav1.GetOptions{})
-			if err != nil {
-				return "", "", err
-			}
+func getTLSCertInfo(kubeClient kubernetes.Interface, namespace string, portConfig *portConfig) (string, string, error) {
+	if portConfig.TLSSecretName == "" {
+		return "", "", fmt.Errorf("cert & key for port %v is not specified", portConfig.Port)
```
jnschaeffer: Can this be `TLS secret name for port %v is not specified`? The error itself is returned with the secret name isn't specified, while the error makes it seem as if the cert and key keys in the secret need to be specified, which is not the case.

cloud/linode/loadbalancers.go
```diff
@@ -526,3 +523,46 @@ func getConnectionThrottle(service *v1.Service) int {
 
 	return connThrottle
 }
+
+const (
```
jnschaeffer: Non-blocking, but can we put these in their own file in the same package?

cloud/linode/loadbalancers.go
```diff
@@ -419,18 +418,33 @@ func (l *loadbalancers) retrieveKubeClient() error {
 	return nil
 }
 
-// getProtocol returns the desired protocol of service.
-func getProtocol(service *v1.Service) (linodego.ConfigProtocol, error) {
-	protocol, ok := service.Annotations[annLinodeProtocol]
-	if !ok {
-		return linodego.ProtocolTCP, nil
+func getPortConfig(service *v1.Service, port int) (*portConfig, error) {
```
jnschaeffer: This function probably shouldn't return a `*portConfig`, since there's no case where it returns nil and doesn't also return an error.

cloud/linode/loadbalancers.go
```diff
@@ -485,26 +483,24 @@ func getNodeInternalIP(node *v1.Node) string {
...
 }
 
-func getTLSCertInfo(kubeClient kubernetes.Interface, tlsAnnotations []*tlsAnnotation, namespace string, port int) (string, string, error) {
-	for _, tlsAnnotation := range tlsAnnotations {
-		if tlsAnnotation.Port == port {
-			secret, err := kubeClient.CoreV1().Secrets(namespace).Get(tlsAnnotation.TLSSecretName, metav1.GetOptions{})
-			if err != nil {
-				return "", "", err
-			}
+func getTLSCertInfo(kubeClient kubernetes.Interface, namespace string, portConfig *portConfig) (string, string, error) {
```
jnschaeffer: As above, this can (and probably should) take a `portConfig`, not a `*portConfig`.

PR #51, opened 2019-12-06
Rebuild all NodeBalancer configs on update

jnschaeffer:
It turns out that according to [the API spec](https://developers.linode.com/api/v4/nodebalancers-node-balancer-id-configs/#post), and contrary to [linodego](https://github.com/linode/linodego/blob/master/nodebalancer_configs.go#L104), NodeBalancer configs do not get populated with nodes at create time; they only get populated with nodes at rebuild time. As a result, LoadBalancer services that were created, then updated with new ports weren't getting nodes associated with those ports.

This pull request resolves that issue by creating configs for ports that don't exist, then rebuilding all configs for all ports.

### DIFF
--- a/cloud/linode/loadbalancers.go
+++ b/cloud/linode/loadbalancers.go
@@ -191,12 +191,10 @@ func (l *loadbalancers) UpdateLoadBalancer(ctx context.Context, clusterName stri
 		if currentNBCfg == nil {
 			createOpts := newNBCfg.GetCreateOptions()
 
-			newNBCfg, err := l.client.CreateNodeBalancerConfig(ctx, lb.ID, createOpts)
+			currentNBCfg, err = l.client.CreateNodeBalancerConfig(ctx, lb.ID, createOpts)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


cloud/linode/loadbalancers.go
```diff
@@ -178,31 +178,32 @@ func (l *loadbalancers) UpdateLoadBalancer(ctx context.Context, clusterName stri
...
-				return fmt.Errorf("[port %d] error rebuilding NodeBalancer config: %v", int(port.Port), err)
-			}
-		} else {
+		// If there's no existing config, create it
+		if currentNBCfg == nil {
 			createOpts := newNBCfg.GetCreateOptions()
-			createOpts.Nodes = newNBNodes
 
-			_, err := l.client.CreateNodeBalancerConfig(ctx, lb.ID, createOpts)
+			newNBCfg, err := l.client.CreateNodeBalancerConfig(ctx, lb.ID, createOpts)
```
aspsk: I believe the `err` is defined already, so you can do `currentNBCfg, err = l.client...` to save two lines below

PR #54, opened 2020-01-09
Added support for Sentry

jnschaeffer:
This commit adds support for Sentry error reporting via a new `sentry` package, as well as error reporting at certain key locations in the CCM. Consumers are expected to only use the `sentry` package, and not the Sentry SDK directly. Functions log errors on failure rather than return error values; as a result, consumers should not assume all captured errors are being sent to Sentry.


phillc:
It feels odd for this to be Sentry specific. Feels like some sort of interface around error logging that can be injected can be refactored to in the future.

There is a TODO still in the code.

phillc:
I feel like some of this can be cleaned up with a 
`defer if err != nil { Sentry... }`

sentry/sentry.go
```diff
@@ -0,0 +1,82 @@
...
+
+	return sentry.SetHubOnContext(ctx, hub)
+}
+
+// getHubFromContext gets the current Sentry hub from the given context. If the context is missing a
+// Sentry hub, this function logs an error and returns nil. If Sentry has not been initialized, this
+// function also returns nil.
+func getHubFromContext(ctx context.Context) *sentry.Hub {
+	if !initialized {
+		glog.V(3).Info("getHubFromContext: Sentry not initialized")
```
asauber: Let's make this V5. A lot of users run with `--v=3` and I would rather not have this clutter up their logs.

cloud/linode/instances.go
```diff
@@ -108,15 +147,24 @@ func (i *instances) CurrentNodeName(_ context.Context, hostname string) (types.N
...
 	}
+
+	sentry.SetTag(ctx, "linode_id", id)
+
 	_, err = linodeByID(ctx, i.client, id)
 	if err == nil {
 		return true, nil
 	}
 
+	// TODO: Should we capture an error here?
```
asauber: since we're not capturing errors for any of the un-exported methods here, yes, let's capture this error

cloud/linode/loadbalancers.go
```diff
@@ -233,6 +267,10 @@ func (l *loadbalancers) deleteUnusedConfigs(ctx context.Context, nbConfigs []lin
 //
 // EnsureLoadBalancerDeleted will not modify service.
 func (l *loadbalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterName string, service *v1.Service) error {
+	ctx = sentry.SetHubOnContext(ctx)
+	sentry.SetTag(ctx, "cluster_name", clusterName)
+	sentry.SetTag(ctx, "service", service.Name)
+
 	_, exists, err := l.GetLoadBalancer(ctx, clusterName, service)
 	if err != nil {
 		return err
```
asauber: this error is worth capturing as well

jnschaeffer: `GetLoadBalancer` already captures errors, since it can also be called on its own, which is why the error isn't captured there. I should document that, though.

cloud/linode/instances.go
jnschaeffer: Done!

cloud/linode/loadbalancers.go
asauber: right, I see that, this looks good then.

PR #69, opened 2020-06-12
Add loadbalancer preserve annotation

Charliekenney23:
This change adds an annotation that can be used on services of `type: LoadBalancer` to ensure that the underlying NodeBalancer is not deleted when the service is.

This is the first of two PRs to enable reusing LoadBalancers.

### DIFF
--- a/cloud/linode/loadbalancers.go
+++ b/cloud/linode/loadbalancers.go
@@ -278,9 +278,7 @@ func (l *loadbalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterNa
 
 	// Don't delete the underlying nodebalancer if the service has the preserve annotation.
 	if preserveRaw, ok := service.Annotations[annLinodeLoadBalancerPreserve]; ok {
-		if preserve, err := strconv.ParseBool(preserveRaw); err != nil {
-			return fmt.Errorf("failed to parse annotation %s value: %s", annLinodeLoadBalancerPreserve, err)
-		} else if preserve {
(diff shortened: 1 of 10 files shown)


cloud/linode/loadbalancers.go
```diff
@@ -274,6 +276,13 @@ func (l *loadbalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterNa
 	sentry.SetTag(ctx, "cluster_name", clusterName)
 	sentry.SetTag(ctx, "service", service.Name)
 
+	// Don't delete the underlying nodebalancer if the service has the preserve annotation.
+	if preserveRaw, ok := service.Annotations[annLinodeLoadBalancerPreserve]; ok {
+		if preserve, err := strconv.ParseBool(preserveRaw); err == nil && preserve {
+			return nil
+		}
```
asauber, line 283: Missing a `return err` branch here.

Charliekenney23: This was intentional. When talking with @jnschaeffer, we thought it'd be better to interpret an invalid boolean as `false` and document it as such. Do you think there is value in throwing an error here instead?

asauber: I had this comment as pending for a while. Sorry for the confusion in chat.

Charliekenney23: I added a note below the annotation lists describing the accepted string representations for `true`, [here](https://github.com/linode/linode-cloud-controller-manager/pull/69/files#diff-04c6e90faac2675aa89e2176d2eec7d8R66).

PR #76, opened 2020-08-21
add nodebalancer-id annotation

Charliekenney23:


cloud/linode/loadbalancers.go
```diff
@@ -316,30 +358,35 @@ func (l *loadbalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterNa
...
+func (l *loadbalancers) getNodeBalancerByID(ctx context.Context, service *v1.Service, id int) (lb *linodego.NodeBalancer, err error) {
+	lb, err = l.client.GetNodeBalancer(ctx, id)
+	if err != nil {
+		if apiErr, ok := err.(*linodego.Error); ok && apiErr.Code == http.StatusNotFound {
+			err = errLbNotFound
+		}
+	} else {
+		klog.Infof("found NodeBalancer (%d) for service %s/%s via %s annotation", id, service.Namespace, service.Name, annLinodeNodeBalancerID)
+	}
+	return
```
sibucan, line 384: It looks like you can drop the `else` if you return early during the error condition, like so:
```suggestion
	lb, err = l.client.GetNodeBalancer(ctx, id)
	if err != nil {
		if apiErr, ok := err.(*linodego.Error); ok && apiErr.Code == http.StatusNotFound {
			return nil, errLbNotFound
		}
	    return nil, err
	}
	klog.Infof("found NodeBalancer (%d) for service %s/%s via %s annotation", id, service.Namespace, service.Name, annLinodeNodeBalancerID)
	return lib, nil
```

cloud/linode/loadbalancers.go
```diff
@@ -67,6 +70,62 @@ func newLoadbalancers(client *linodego.Client, zone string) cloudprovider.LoadBa
 	return &loadbalancers{client: client, zone: zone}
 }
 
+func (l *loadbalancers) getNodeBalancerForService(ctx context.Context, service *v1.Service) (*linodego.NodeBalancer, error) {
+	rawID, _ := getServiceAnnotation(service, annLinodeNodeBalancerID)
+	id, idErr := strconv.Atoi(rawID)
+	hasIDAnn := idErr == nil && id != 0
+
+	if hasIDAnn {
```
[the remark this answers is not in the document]
Charliekenney23: The reason why I didn't exit early is so that if the NodeBalancer specified by the ID does not exist, it falls back to looking for the NodeBalancer by the most recent status.

e2e/test/framework/loadbalancer_suit.go
```diff
@@ -3,6 +3,7 @@ package framework
 import (
```
asauber, line 3: It would be nice to rename this file `loadbalancer_suite.go` in this PR if you don't mind.

README.md
```diff
@@ -52,7 +52,8 @@ Annotation (Suffix) | Values | Default | Description
 `check-timeout` | int (1-30) | | Duration, in seconds, to wait for a health check to succeed before considering it a failure
 `check-attempts` | int (1-30) | | Number of health check failures necessary to remove a back-end from the service
 `check-passive` | [bool](#annotation-bool-values) | `false` | When `true`, `5xx` status codes will cause the health check to fail
-`preserve` | [bool](#annotation-bool-values) | `false` | When `true`, deleting a `LoadBalancer` service does not delete the underlying NodeBalancer
+`preserve` | [bool](#annotation-bool-values) | `false` | When `true`, deleting a `LoadBalancer` service does not delete the underlying NodeBalancer. Replaced NodeBalancers will also not be deleted
```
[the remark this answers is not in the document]
Charliekenney23: Replacing here means when you annotate an existing LoadBalancer service to use a different NodeBalancer. I will write something more to that effect for clarity

cloud/linode/loadbalancers.go
```diff
@@ -243,15 +308,44 @@ func (l *loadbalancers) UpdateLoadBalancer(ctx context.Context, clusterName stri
...
+	}
+
+	nb, err := l.getNodeBalancerForService(ctx, serviceWithStatus)
+	if err != nil {
+		sentry.CaptureError(ctx, err)
+		return err
+	}
+
+	if !l.shouldPreserveNodeBalancer(service) {
+		if err := l.cleanupOldNodeBalancer(ctx, service); err != nil {
```
asauber, line 340: This should only be called in the case where the `nodebalancer-id` annotation does not match the existing NodeBalancer ID. We don't want users to find that when they change a parameter for their Service, such as the port, that they get an entirely new NodeBalancer IP.

asauber (changes requested):

cloud/linode/loadbalancers.go
asauber: Nevermind, the following handles this case :)

https://github.com/linode/linode-cloud-controller-manager/pull/76/files#diff-3ec7caaf647e4dffd0640f14c7410c02R153-R155

asauber (changes requested):
Tested the following scenarios and they all work as expected!

- Adding the `nodebalancer-id` annotation to an existing Service
- Adding `preserve` to a Service that also uses `nodebalancer-id`
- Using `nodebalancer-id` with a new Service
- Upgrading from the previous version of the CCM to this version, and then making edits to the service

Perhaps the best thing about this update is that you can now relabel the NodeBalancers at will without any unexpected side effects!

Two points of feedback:

- New NodeBalancers should receive a unique and LKE-related label. Perhaps `lke<clusterID>-<lower 48 bits of UnixNano() in hex>`
- If a Service is created with `nodebalancer-id`, but that NodeBalancer doesn't exist, then a new NodeBalancer is created. I think this should instead be logged as an error and not create any NodeBalancer.

asauber:
Looks good to me. Thanks for the updates!

Seen when updating an existing Service or creating an new service with an invalid ID:

```
service.beta.kubernetes.io/linode-loadbalancer-nodebalancer-id annotation points to a NodeBalancer that does not exist: LoadBalancer (2340981) not found for service
```

asauber (approved):

PR #87, opened 2020-12-16
allow inter-service communication from within cluster via external LB

### LINKED ISSUE #74: LoadBalancer Services cannot discover the true ClientIP
X-Forwarded-For, X-Real-IP, ProxyProtocol and friends currently see only an IP in the range 192.168.255.0/24

Charliekenney23:
Closes #74

This changes the LoadBalancer status we return from `EnsureLoadBalancer` and `GetLoadBalancer` to only include the NodeBalancer hostname and not the IPv4 address. Omitting the IP ensures kube-proxy does not insert rules to hijack requests to the external LB IP from within the cluster to resolve to the cluster IP for the underlying pod. This proxying caused issues for users expecting the NodeBalancer to include proxy-protocol headers. See #74.

Due to LoadBalancer statuses not including the LB IPv4 address, LB discovery will now be done using the hostname.

### DIFF
--- a/cloud/linode/loadbalancers.go
+++ b/cloud/linode/loadbalancers.go
@@ -120,11 +120,13 @@ func (l *loadbalancers) getLatestServiceLoadBalancerStatus(ctx context.Context, 
 
 // getNodeBalancerByStatus attempts to get the NodeBalancer from the IPv4 specified in the
 // most recent LoadBalancer status.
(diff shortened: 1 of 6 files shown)


phillc (approved):
I'd get 1 more approver.

sibucan (approved):

PR #92, opened 2021-03-03
add loadbalancer-hostname-only-ingress annotation

Charliekenney23:
This annotation will allow users to use short-circuit adding the underlying NodeBalancer's IP to the LoadBalancerStatus (see #87). This is now an opt-in feature as it's much quicker to expose the IPv4 as an interface for the service than waiting for the DNS A record to be propagated for the hostname.

### DIFF
--- a/cloud/linode/loadbalancers.go
+++ b/cloud/linode/loadbalancers.go
@@ -124,11 +124,11 @@ func (l *loadbalancers) getLatestServiceLoadBalancerStatus(ctx context.Context, 
 // most recent LoadBalancer status.
 func (l *loadbalancers) getNodeBalancerByStatus(ctx context.Context, service *v1.Service) (nb *linodego.NodeBalancer, err error) {
 	for _, ingress := range service.Status.LoadBalancer.Ingress {
-		if ingress.Hostname != "" {
-			return l.getNodeBalancerByHostname(ctx, service, ingress.Hostname)
-		}
 		if ingress.IP != "" {
 			return l.getNodeBalancerByIPv4(ctx, service, ingress.IP)
+		}
+		if ingress.Hostname != "" {
+			return l.getNodeBalancerByHostname(ctx, service, ingress.Hostname)
 		}
 	}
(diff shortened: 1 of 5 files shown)


cloud/linode/loadbalancers.go
```diff
@@ -769,11 +766,16 @@ func getConnectionThrottle(service *v1.Service) int {
...
 }
 
-func makeLoadBalancerStatus(nb *linodego.NodeBalancer) *v1.LoadBalancerStatus {
+func makeLoadBalancerStatus(service *v1.Service, nb *linodego.NodeBalancer) *v1.LoadBalancerStatus {
+	ingress := v1.LoadBalancerIngress{
+		Hostname: *nb.Hostname,
+	}
+	if !getServiceBoolAnnotation(service, annLinodeHostnameOnlyIngress) {
+		ingress.IP = *nb.IPv4
+	}
```
sibucan, line 775: It seems strange that the annotation for a NB hostname on the svc is opt-in, but the logic in here implies the other way around, that the default is to return the NB's hostname unless a specific annotation is set. Do you think it could maybe it read like:
```suggestion
	ingress := v1.LoadBalancerIngress{
		Hostname: *nb.IPv4,
	}
	if getServiceBoolAnnotation(service, annLinodeHostnameOnlyIngress) {
		ingress.IP = *nb.Hostname
	}
```
... or am I missing a subtle bug by doing this?

sibucan (approved):
Just a small comment about how the logic for choosing whether to show the NB IP or the hostname based on the annotation:

cloud/linode/loadbalancers.go
Charliekenney23: Before the kube-proxy workaround, we returned both the IP and hostname. Now If the new annotation is present we only want the hostname. Otherwise, we want the old default of having the IP and hostname

g1bl3t (approved):
This looks good. made a small suggestion for the docs just to clarify a bit

PR #98, opened 2021-08-03
Hardcode --authentication-skip-lookup=true to short-circuit ConfigMap auth

### LINKED ISSUE #91: Add --authentication-skip-lookup=true flag
The example deployment manifest for the Linode CCM uses the cluster-admin role. When deploying the CCM with a limited clusterrole, the CCM fails to start if it doesn't have access to `get` on `configmaps`:
```
W0222 19:11:17.239825       1 requestheader_controller.go:193] Unable to get configmap/extension-apiserver-authentication in kube-system.  Usually fixed by 'kubectl create rolebinding -n kube-system ROLEBINDING_NAME --role=extension-apiserver-authentication-reader --serviceaccount=YOUR_NS:YOUR_SA'
unable to load configmap based request-header-client-ca-file: configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:kube-system:default" cannot get resource "configmaps" in API group "" in the namespace "kube-system"
```
This seems to be related to upstream issue, see: https://github.com/kubernetes/cloud-provider/issues/34

It looks like Digital Ocean ran into a similar problem after updating their K8s deps (https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/217). Possible short-term solutions could be to add the `--authentication-skip-lookup=true` flag to the deployment example manifest, or to hardcode this option in the binary

Charliekenney23:
Resolves #91

This change hardcodes some flags. Setting `--authentication-skip-lookup=true` short circuits configmap lookup for authentication (no longer needed), allowing for more fine-grained permissions. Also hardcodes --cloud-provider to "linode".

The manifest template has also been updated to use a restricted clusterrolebinding.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,6 @@ run: build
 	dist/linode-cloud-controller-manager \
 		--logtostderr=true \
 		--stderrthreshold=INFO \
-		--cloud-provider=linode \
 		--kubeconfig=${KUBECONFIG}
 
 .PHONY: run-debug
@@ -77,7 +76,6 @@ run-debug: build
 	dist/linode-cloud-controller-manager \
 		--logtostderr=true \
 		--stderrthreshold=INFO \
-		--cloud-provider=linode \
 		--kubeconfig=${KUBECONFIG} \
 		--linodego-debug
 
--- a/deploy/ccm-linode-template.yaml
+++ b/deploy/ccm-linode-template.yaml
@@ -101,7 +101,6 @@ spec:
           imagePullPolicy: Always
           name: ccm-linode
           args:
-          - --cloud-provider=linode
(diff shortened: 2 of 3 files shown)


sibucan:
Since we're hardcoding the `--cloud-provider` flag to "linode" within the application, is the flag necessary in the template? It wouldn't have any effect on the behavior of the application: https://github.com/linode/linode-cloud-controller-manager/blob/e41351398272a549363638fb3ea5e7d82acc5b0f/deploy/ccm-linode-template.yaml#L75

sibucan (approved):
👍 Tested, the CCM now has default values set for the `cloud-provider` and `authentication-skip-lookup` flags.

PR #99, opened 2021-09-13
Fix bug that caused generic API errors via InstanceExistsByProviderID to delete nodes

Charliekenney23:
These changes refactor the implementation of `InstanceExistsByProviderID` to only return false without error, and thereby causing the corresponding node to be deleted, when the linodego client returns a "Not Found" error. Any other errors will be propagated, and the call to `InstanceExistsByProviderID` will be retried.

The actual bug fix is in [this commit](https://github.com/linode/linode-cloud-controller-manager/commit/8d70a47ae286d8deb3b9bbe7667f074551635fe9).

The codebase needed to be refactored to test for this bug:
- adopt gomock and generate linodego client mock - a34db90
- break out shared helpers, replace `*linodego.Client` with `linodeClient`, etc. - 5b959a7
- Remove unneeded test code for helpers in `fakeAPI` - b0111b2

Eventually, `fakeAPI` should be removed in favor of the `linodeClient` mock. The loadbalancers test code can be updated later on.


frenchtoasters (approved):

sibucan (approved):
Tested the changes by mocking API responses. Before, on any error that the CCM received while the node was `NotReady`, the CCM would delete the node with the following logs:
```
I0915 18:58:13.065706       1 node_lifecycle_controller.go:156] deleting node since it is no longer present in cloud provider: test-119-119-machine2
I0915 18:58:13.066067       1 event.go:291] "Event occurred" object="test-node2" kind="Node" apiVersion="" type="Normal" reason="Deleting node ttest-node2 because it does not exist in the cloud provider" message="Node test-node2 event: DeletingNode"
```

After the changes in this PR, I see the following in the logs:
```
E0915 18:49:37.064812       1 node_lifecycle_controller.go:149] error checking if node test-node2 exists: [502] Bad gateway
```

And the node is not deleted by the CCM, as desired. Only after the API returns a 404 will the node be deleted as expected.

g1bl3t (approved):
LGTM, thanks for the quick fix on this!

PR #108, opened 2022-04-14
Update k8s dependencies to v1.21.0

### LINKED ISSUE #107: CCM needs agent/worker nodes to create the automatic loadbalancer for a service/ingress
I suppose the easiest way to reproduce, would be to create a single node cluster with K3s and make it a master. Apply the generated ccm-linode.yaml following the howto and then try to LoadBalance a service. It will fail. Add an agent node to the cluster and then it will succeed.

If I'm correct about this, it is limiting our options.
For example, I have setup a HA K3s for rancher and planning to use it, to create more clusters. The 3 control plane nodes should be able to do the whole job. There should be no need to add agent nodes, because you would also need at least 2.

I thought I'd just reach out and see if this is something I'm doing wrong and there is a label, or an annotation I'm missing, that might solve my problem.

sibucan:
### General:

### Pull Request Guidelines:

1. [X] Does your submission pass tests?
1. [ ] Have you added tests? 
1. [X] Have you linked your PR to an [open issue](https://blog.github.com/2013-05-14-closing-issues-via-pull-requests/)

This PR upgrades all Kubernetes-related dependencies from v1.19.2 to v1.21.0. This includes some changes on how the CCM command is initialized, which is explained in https://github.com/kubernetes/kubernetes/blob/v1.21.0/cmd/cloud-controller-manager/main.go and https://github.com/kubernetes/cloud-provider/blob/master/sample/README.md

Now that the Linode CCM is using v1.21.0 k8s dependencies, it should fix #107


g1bl3t (approved):
tested and working 👍

Charliekenney23 (approved):
LGTM

PR #118, opened 2023-02-23
Update linodego to v1

okokes-akamai:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


okokes-akamai:
Once #126 gets merged, rebase this and add the following to `getNodeBalancerByIPv4` instead of the hardcoded filter (this constructor is new to v1).

```go
f := &linodego.Filter{}
f.AddField(linodego.Eq, "ipv4", ipv4)
filter, err := f.MarshalJSON()
if err != nil {
	return nil, err
}
```

okokes-akamai:
Bumped linodego in e2e deps as well, just haven't had a chance to test it as I have issues with my environment.

luthermonson:
@okokes-akamai if this is still good to go then let's get it merged. could you also bump go to 1.20 while youre in the go.mod?

go.mod
```diff
@@ -6,10 +6,10 @@ require (
 	github.com/appscode/go v0.0.0-20200323182826-54e98e09185a
 	github.com/getsentry/sentry-go v0.4.0
 	github.com/golang/mock v1.6.0
-	github.com/linode/linodego v0.32.2
+	github.com/linode/linodego v1.14.1
```
luthermonson, line 9: think you could go all the way to v1.25.0 now?

okokes-akamai:
> @okokes-akamai if this is still good to go then let's get it merged. could you also bump go to 1.20 while youre in the go.mod?

This is definitely not ready to be merged, because it has implications for our deployment model due to the linodego URL handling. So we cannot just merge it, because then we'd have to rework our internal systems accordingly (and would thus be blocked from developing CCM until that happens).

rammanoj:
This is being taken care at #143. @okokes-akamai I believe this should be good to close.

PR #123, opened 2023-03-03
Use InstancesV2 to limit API calls (and simplify the code)

okokes-akamai:
## Rationale

There's a new-ish interface, `InstancesV2`, which is a) a lot simpler than `Instances`, b) calls the cloud API much less often, so we end up with a lot less code that should be also more efficient.

I tried to split the implementation into individual commits while keeping the tests green, so at first I implemented both interfaces at once and then deleted the old one once I had tests in place for the new one.

I also removed the Zones implementation, because it's no longer called within `InstancesV2` - I did move its tests into `instances_test.go`, so its functionality is preserved.

### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 


cloud/linode/instances.go
```diff
@@ -29,176 +27,76 @@ func (e instanceNoIPAddressesError) Error() string {
...
-
-func (i *instances) InstanceShutdownByProviderID(ctx context.Context, providerID string) (bool, error) {
-	ctx = sentry.SetHubOnContext(ctx)
-	sentry.SetTag(ctx, "provider_id", providerID)
-
-	id, err := parseProviderID(providerID)
-	if err != nil {
-		sentry.CaptureError(ctx, err)
-		return false, err
+	// note that Zone is omitted as it's not a thing in Linode
```
mrogers950, line 93: With Zone being removed, this comment mostly becomes historical.

okokes-akamai: `zones.go` is begin removed, but its functionality is folded into this big metadata method. There really isn't a difference in functionality between master and this PR, the original zones.go code also omitted Zone (as in the struct field), I'm just stating it here explicitly, because it's the only struct field I'm not filling in - so I'm stating, for _future_ reference, that the omission is intentional.

git-it127:
I ran the [e2e tests](https://github.com/linode/linode-cloud-controller-manager/tree/master/e2e) here and they all fail.

I believe this has something to do with the nodebalancers being changed from the name `ccm-<hash>` to `balancer<hash>`. We likely want to keep the naming consistent for compatibility reasons.

mrogers950 (approved):
I checked this out alongside https://github.com/linode/linode-cloud-controller-manager/pull/126 and ran a quick manual test adding a node to a cluster, it worked fine. No issues with the code from my view.

cloud/linode/instances.go
```diff
@@ -29,176 +29,89 @@ func (e instanceNoIPAddressesError) Error() string {
...
-	addresses, err := i.nodeAddresses(ctx, linode)
-	if err != nil {
-		sentry.CaptureError(ctx, err)
-		return nil, err
-	}
-
-	return addresses, nil
-}
+func (i *instances) lookupLinode(ctx context.Context, node *v1.Node) (*linodego.Instance, error) {
+	providerID := node.Spec.ProviderID
```
git-it127, line 33: I believe the sentry context needs set here.
```suggestion
	ctx = sentry.SetHubOnContext(ctx)
	providerID := node.Spec.ProviderID
```

cloud/linode/instances.go
```diff
@@ -29,176 +29,89 @@ func (e instanceNoIPAddressesError) Error() string {
...
+		return linodeByID(ctx, i.client, id)
 	}
-	// Allow Nodes to not have an InternalIP, if this proves problematic this will be reverted
 
-	return addresses, nil
+	return linodeByName(ctx, i.client, nodeName)
 }
 
-func (i *instances) InstanceID(ctx context.Context, nodeName types.NodeName) (string, error) {
+func (i *instances) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
```
git-it127, line 53: Can we rename this function to be in the form of a true / false question since this is returning a bool.

I like things that start with is, has, can, was.

Something like
```suggestion
func (i *instances) HasInstance(ctx context.Context, node *v1.Node) (bool, error) {
```

cloud/linode/instances.go
```diff
@@ -29,176 +29,89 @@ func (e instanceNoIPAddressesError) Error() string {
...
-	if err != nil {
-		sentry.CaptureError(ctx, err)
-		return "", err
-	}
-	return linode.Type, nil
+	return true, nil
 }
 
-func (i *instances) InstanceTypeByProviderID(ctx context.Context, providerID string) (string, error) {
+func (i *instances) InstanceShutdown(ctx context.Context, node *v1.Node) (bool, error) {
```
git-it127, line 66: Same here can we rename to a true / false question.

Maybe something like this
```suggestion
func (i *instances) IsInstanceShutdown(ctx context.Context, node *v1.Node) (bool, error) {
```

git-it127 (changes requested):
Overall this looks pretty good. I have a few suggestions.

cloud/linode/instances.go
okokes-akamai: Note that all invocations of `lookupLinode` are preceded by the `SetHubOnContext`, so I don't think this call is needed. I'm not even sure it'd work: It's called in the parent, because `SetHubOnContext` returns a copy of said context and `lookupLinode` then has no way of returning that context to the parent.

cloud/linode/instances.go
okokes-akamai: We cannot rename this method, because then we wouldn't implement the [upstream interface](https://pkg.go.dev/k8s.io/cloud-provider#InstancesV2).

cloud/linode/instances.go
okokes-akamai: Addressed above.

scmeggitt:
After discussing, we decided it does not make sense to upgrade the version right now.

scmeggitt (approved):

git-it127 (approved):
LGTM

PR #131, opened 2023-05-17
Cache node listings instead of random lookups

okokes-akamai:
### Rationale:

Since we call the individual "get instance" API call very frequently, we could instead call a single "list instances" once and reuse that for all the Linode lookups. This is essentially a client-side cache.

We won't reduce traffic between us and the Linode API (we'll probably increase it a bit), but we'll cut down the number of calls quite dramatically.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/cloud/linode/instances.go
+++ b/cloud/linode/instances.go
@@ -21,7 +21,9 @@ type nodeCache struct {
 	ttl        time.Duration
 }
 
-func (nc *nodeCache) refresh(ctx context.Context, client Client) error {
+// refreshInstances conditionally loads all instances from the Linode API and caches them.
+// It does not refresh if the last update happened less than `nodeCache.ttl` ago.
+func (nc *nodeCache) refreshInstances(ctx context.Context, client Client) error {
 	nc.Lock()
 	defer nc.Unlock()
 
@@ -52,7 +54,7 @@ type instances struct {
 func newInstances(client Client) cloudprovider.InstancesV2 {
 	return &instances{client, &nodeCache{
 		nodes: make(map[int]*linodego.Instance),
-		ttl:   90 * time.Second,
+		ttl:   15 * time.Second,
 	}}
 }
 
@@ -87,7 +89,7 @@ func (i *instances) linodeByID(id int) (*linodego.Instance, error) {
 }
 
 func (i *instances) lookupLinode(ctx context.Context, node *v1.Node) (*linodego.Instance, error) {
-	if err := i.nodeCache.refresh(ctx, i.client); err != nil {
+	if err := i.nodeCache.refreshInstances(ctx, i.client); err != nil {
 		return nil, err
 	}
 
--- a/cloud/linode/instances.go
+++ b/cloud/linode/instances.go
@@ -29,11 +29,11 @@ func (nc *nodeCache) refresh(ctx context.Context, client Client) error {
 		return nil
 	}
 
-	nc.nodes = make(map[int]*linodego.Instance)
 	instances, err := client.ListInstances(ctx, nil)
 	if err != nil {
 		return err
 	}
+	nc.nodes = make(map[int]*linodego.Instance)
 	for _, instance := range instances {
 		instance := instance
 		nc.nodes[instance.ID] = &instance
--- a/cloud/linode/instances.go
+++ b/cloud/linode/instances.go
@@ -3,6 +3,7 @@ package linode
 import (
 	"context"
 	"fmt"
+	"os"
 	"strconv"
 	"sync"
 	"time"
@@ -52,9 +53,17 @@ type instances struct {
 }
 
 func newInstances(client Client) cloudprovider.InstancesV2 {
+	var timeout int
+	if raw, ok := os.LookupEnv("LINODE_INSTANCE_CACHE_TTL"); ok {
+		timeout, _ = strconv.Atoi(raw)
+	}
+	if timeout == 0 {
+		timeout = 15
+	}
+
 	return &instances{client, &nodeCache{
 		nodes: make(map[int]*linodego.Instance),
-		ttl:   15 * time.Second,
+		ttl:   time.Duration(timeout) * time.Second,
 	}}
 }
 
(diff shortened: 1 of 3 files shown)


cloud/linode/instances_test.go
```diff
@@ -33,50 +31,36 @@ func TestInstanceExists(t *testing.T) {
...
-		expectedErr := errors.New("some error")
-		client.EXPECT().GetInstance(gomock.Any(), 123).Times(1).Return(nil, expectedErr)
-
-		exists, err := instances.InstanceExists(ctx, node)
-		assert.ErrorIs(t, err, expectedErr)
-		assert.False(t, exists)
-	})
 
 	t.Run("should return false if linode does not exist (by providerID)", func(t *testing.T) {
+		instances := newInstances(client)
```
okokes-akamai, line 36: Had to hoist in the `newInstances` initialisation, otherwise we'd get test pollution due to a shared cache.

cloud/linode/instances_test.go
```diff
@@ -33,50 +31,36 @@ func TestInstanceExists(t *testing.T) {
...
 		exists, err := instances.InstanceExists(ctx, node)
 		assert.NoError(t, err)
 		assert.False(t, exists)
 	})
 
 	t.Run("should return false if linode does not exist (by name)", func(t *testing.T) {
+		instances := newInstances(client)
 		name := "some-name"
 		node := nodeWithName(name)
-		notFound := &linodego.Error{
```
okokes-akamai, line 62: The nice thing about this implementation is that we no longer use http-specific errors where in practice we don't care about HTTP issues, we just want to retrieve some data.

cloud/linode/instances.go
```diff
@@ -13,12 +14,46 @@ import (
 	cloudprovider "k8s.io/cloud-provider"
 )
 
+type nodeCache struct {
+	sync.RWMutex
+	nodes      map[int]*linodego.Instance
+	lastUpdate time.Time
+	ttl        time.Duration
```
scmeggitt, line 21: should this be a flag we can change if we need to?

cloud/linode/instances.go
```diff
@@ -13,12 +14,46 @@ import (
...
+}
+
+func (nc *nodeCache) refresh(ctx context.Context, client Client) error {
```
scmeggitt, line 24: would it make sense to name this `nodeCacheRefresh` or `instanceRefresh`?

scmeggitt:
Should we add a test that adds multiple nodes to test that it works as well?

cloud/linode/instances.go
```diff
@@ -29,7 +64,33 @@ func (e instanceNoIPAddressesError) Error() string {
...
+	defer i.nodeCache.RUnlock()
+	instance, ok := i.nodeCache.nodes[id]
+	if !ok {
+		return nil, cloudprovider.InstanceNotFound
+	}
+	return instance, nil
+}
+
 func (i *instances) lookupLinode(ctx context.Context, node *v1.Node) (*linodego.Instance, error) {
+	if err := i.nodeCache.refresh(ctx, i.client); err != nil {
```
scmeggitt, line 90: So if a new node gets added, it will take 90 seconds for the nodeCache to update with the new node? That seems a little long, is there a reason for that number?

cloud/linode/instances.go
okokes-akamai: Yeah, could be. It's just that the init for this (and all its parent inits) is not via main.go, it goes through `init()`, so it'd have to be an env variable.

cloud/linode/instances.go
okokes-akamai: Note that it's a method on `nodeCache`, so the (only) call site is literally `instances.nodeCache.refresh`, so including `nodeCache` or `instance` in the name would only duplicate prior mentions of these.

Though come to think of it, we do shorten `instances` to `i` when calling it, so we could make it `refreshInstances` and the call site would then be `i.nodeCache.refreshInstances(...)`.

cloud/linode/instances.go
okokes-akamai: The duration is fairly arbitrary. I wanted all the subsequent calls to hit the cache, but they do happen in succession, so we could shorten this quite a bit. I thought we were constrained by the 5-minute refresh intervals anyway, so 90 seconds seemed like a good middle ground. I'll shorten it to 15 seconds and I'll grep logs to see how much our calls to the API are bundled together.

cloud/linode/instances.go
scmeggitt: I think it might be worth doing just in the off chance we need to tune this, then it won't be a pain. We also might want to have different ttl for the child clusters vs the CPC.

cloud/linode/instances.go
scmeggitt: The 5 minute interval is only for updating the metadata, but when nodes are created or shutdown, that 5 minute interval does not come into play. The node monitor period which is involved in checking if a node is up is being increased from 5 seconds to 10 seconds which I think will interact with the cache more. So 15 seconds would work, or we could do 30.

cloud/linode/instances.go
```diff
@@ -29,7 +75,33 @@ func (e instanceNoIPAddressesError) Error() string {
...
+func (i *instances) linodeByName(nodeName types.NodeName) (*linodego.Instance, error) {
+	i.nodeCache.RLock()
+	defer i.nodeCache.RUnlock()
+	for _, node := range i.nodeCache.nodes {
+		if node.Label == string(nodeName) {
+			return node, nil
+		}
+	}
+
+	return nil, cloudprovider.InstanceNotFound
```
mrogers950, line 87: I wonder if now we should still handle this as `cloudprovider.InstanceNotFound` since technically it's not found from our own cache, and not the provider.. It may be all the same either way.

okokes-akamai: I see the cache as an implementation detail - the fact that we're not interrogating the API live is just our internal logic.

What would you suggest as an alternative? `errors.New("instance not found")`, which would suggest we didn't find it ourselves, not that it's not present upstream within the cloud provider?

scmeggitt (approved):
Looks good

mrogers950 (approved):
LGTM

PR #143, opened 2023-11-27
Attach firewall when creating nodebalancer

### LINKED ISSUE #142: Support Firewall attachments to nodebalancers
## General:

----

## Feature Requests:
* [ ] Have you explained your rationale for why this feature is needed? 
* [x] Have you offered a proposed implementation/solution? 

Support attaching a firewall to a node-balancer. The preferred approach would be to attach this by setting an annotation on the service. The ccm should attach the firewall specified in the annotation `service.beta.kubernetes.io/linode-loadbalancer-firewall-id` .

rammanoj:
Fixes #142 

**Changes:**
- Upgrade Linodego dependency to the latest available version v1.25.0 and fix the existing unit-test failures due to the changes.
- Add support to attach firewall to node balancer when creating a node balancer. 
  - Add corresponding unit-tests:
     - Create NodeBalancer without a firewall
     - Create NodeBalancer with an invalid firewall ID
     - Create NodeBalaner with a valid firewall ID

### General:

### Pull Request Guidelines:

1. [X] Does your submission pass tests?
1. [x] Have you added tests? 
1. [X] Have you linked your PR to an [open issue](https://blog.github.com/2013-05-14-closing-issues-via-pull-requests/)


cloud/linode/loadbalancers.go
```diff
@@ -510,7 +511,7 @@ func (l *loadbalancers) getLoadbalancerTags(ctx context.Context, service *v1.Ser
 	return []string{}
 }
 
-func (l *loadbalancers) createNodeBalancer(ctx context.Context, clusterName string, service *v1.Service, configs []*linodego.NodeBalancerConfigCreateOptions) (lb *linodego.NodeBalancer, err error) {
+func (l *loadbalancers) createNodeBalancer(ctx context.Context, clusterName string, service *v1.Service, configs []*linodego.NodeBalancerConfigCreateOptions, firewallId int) (lb *linodego.NodeBalancer, err error) {
```
schinmai-akamai, line 514: does the firewallID need to be passed in here? 
Can we look it up like we do the Tags?

schinmai-akamai:
I think you have to run `go mod tidy` to fix the lint issues

Makefile
```diff
@@ -17,7 +17,7 @@ codegen:
 .PHONY: lint
 lint:
 	docker run --rm -v "$(shell pwd):/var/work:ro" -w /var/work \
-		golangci/golangci-lint:v1.44.0 golangci-lint run -v --timeout=5m
+		golangci/golangci-lint:v1.46.0 golangci-lint run -v --timeout=5m
```
okokes-akamai, line 20: If we're updating golangci-lint (why are we doing this in this PR?) then maybe bump it all the way to 1.55.2 or whatever is the latest.

cloud/linode/fake_linode_test.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/linode/linodego"
 )
 
+const apiVersion = "v4"
```
okokes-akamai, line 20: Looks like you're doing two PRs in one - updating linodego and this firewall attachment stuff - can you then clean up the PRs, please? I was under the impression we'd do these two PRs in succession, but since you're doing both, it'd be good for this to be cleaned up.

cloud/linode/loadbalancers.go
```diff
@@ -510,18 +511,40 @@ func (l *loadbalancers) getLoadbalancerTags(ctx context.Context, service *v1.Ser
...
+
 	createOpts := linodego.NodeBalancerCreateOptions{
 		Label:              &label,
 		Region:             l.zone,
 		ClientConnThrottle: &connThrottle,
 		Configs:            configs,
 		Tags:               tags,
 	}
+
+	if firewallID != -1 {
```
okokes-akamai, line 545: I'm not a huge fan of sentinel values. For this to be what I'd expect in idiomatic Go code, it'd have to be either `value, ok, error` as return values of the `getLoadBalancerFirewallID` method. Or the error should be comparable to a "firewall not found" error, so that we know if the return value is usable or not.

okokes-akamai: I'd maybe go as far as inlining the getLoadBalancerFirewallID method here, it's simple enough and it makes error handling more direct.

cloud/linode/loadbalancers.go
```diff
@@ -510,18 +511,40 @@ func (l *loadbalancers) getLoadbalancerTags(ctx context.Context, service *v1.Ser
...
+	}
+
+	firewallID, err := strconv.Atoi(fwid)
+	if err != nil {
+		return -1, err
+	}
+	return firewallID, nil
+}
+
 func (l *loadbalancers) createNodeBalancer(ctx context.Context, clusterName string, service *v1.Service, configs []*linodego.NodeBalancerConfigCreateOptions) (lb *linodego.NodeBalancer, err error) {
```
okokes-akamai, line 527: Unit tests are missing for this flow.

cloud/linode/loadbalancers.go
```diff
@@ -510,18 +511,40 @@ func (l *loadbalancers) getLoadbalancerTags(ctx context.Context, service *v1.Ser
...
 		Region:             l.zone,
 		ClientConnThrottle: &connThrottle,
 		Configs:            configs,
 		Tags:               tags,
 	}
+
+	if firewallID != -1 {
+		createOpts.FirewallID = firewallID
+	}
 	return l.client.CreateNodeBalancer(ctx, createOpts)
```
okokes-akamai, line 548: We're only doing the `Read`/`Create` paths of all the CRUD operations we'd want to support. I understand there are limitations to what linodego supports at this point, but I'm not sure we should be merging in something that doesn't even support modification or deletion (!).

okokes-akamai (changes requested):

go.mod
```diff
@@ -6,10 +6,9 @@ require (
 	github.com/appscode/go v0.0.0-20200323182826-54e98e09185a
 	github.com/getsentry/sentry-go v0.4.0
 	github.com/golang/mock v1.6.0
-	github.com/linode/linodego v0.32.2
-	github.com/pkg/errors v0.9.1
+	github.com/linode/linodego v1.25.0
```
mabojars, line 9: Linodego is upgraded significantly here, but not in [e2e/go.mod](https://github.com/rammanoj/linode-cloud-controller-manager/blob/56d93e8b98f38e30c94c2820b36fcf30efaa15d8/e2e/go.mod#L9C17-L9C17). Is there any reason for this mismatch?

Makefile
rammanoj: This is not an intended change for this PR but the upgrade of linodego lead causing https://github.com/golangci/golangci-lint-action/issues/448 issue on linodego package. I had to upgrade it to latest compatible version for it to work. 

I already tried upgrading it to the latest version (1.55.2). In this case, it's flagging multiple issues in the existing code (not the one being edited on the PR). That's the reason, I have upgraded it to the lowest compatible version.

I guess upgrading it to 1.55.2 might require a PR of it's own.

cloud/linode/loadbalancers.go
rammanoj: @okokes-akamai, It's not a limitation from the linodego. It's a limitation of the linode API itself. It doesn't yet support updating/deleting the firewalls on the nodeBalancer. https://github.com/linode/linode-api-docs/pull/863/files contains changes to linode API. Since upgrades and deletes are not supported yet, this feature could probably be called an alpha/something related in the documentation.

go.mod
rammanoj: I missed upgrading it. I'll work on it. Thanks!

rammanoj: The existing e2e tests are already breaking! When I tried executing them without any changes from existing PR, these are breaking! The terraform is not able to find `python` on one of the linodes!
PFA image and the detailed log of the tests.
image
[log.txt](https://github.com/linode/linode-cloud-controller-manager/files/13503877/log.txt)

Since these are already breaking. Fixing these tests and upgrading the linodego in it could be taken on a separate PR.

okokes-akamai:
I have one last comment about testing, but otherwise it looks good.

The one big problem with this is that it only supports creation, which is just... insufficient. I understand we're limited by the Linode API, but given this, we should really make sure this is super alpha, not something that should be used beyond some testing. (cc @schinmai-akamai)

okokes-akamai:
Turns out it should be possible to remove firewalls from nodebalancer by using [DeleteFirewallDevice](https://pkg.go.dev/github.com/linode/linodego#Client.DeleteFirewallDevice) (thanks, Tarun, for pointing this out). So if we're going the creation route, we should also incorporate this (if possible).

rammanoj:
> Turns out it should be possible to remove firewalls from nodebalancer by using [DeleteFirewallDevice](https://pkg.go.dev/github.com/linode/linodego#Client.DeleteFirewallDevice) (thanks, Tarun, for pointing this out). So if we're going the creation route, we should also incorporate this (if possible).

Yup, the devices on the firewalls could be updated or deleted. I will be incorporating these changes as well!

cloud/linode/loadbalancers_test.go
```diff
@@ -233,11 +241,24 @@ func testCreateNodeBalancer(t *testing.T, client *linodego.Client, _ *fakeAPI) {
 		},
 	}
 
+	var errExpected error
```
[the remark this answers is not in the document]
rammanoj: @okokes-akamai Thanks for your comments, I tried table tests by creating a function named `testCreateNodeBalancerOperations`. It included all the newly added 3 tests as table tests. But this is creating issues with the client! Every test requires a new instance of the client. Reusing the same instance of client's returning 4 configs when `ListNodeBalancerConfigs` is called! 

So, I've taken the easier approach and returned the error from `testCreateNodeBalancer` and compared the error in each of the individualized tests.

rammanoj:
> Turns out it should be possible to remove firewalls from nodebalancer by using [DeleteFirewallDevice](https://pkg.go.dev/github.com/linode/linodego#Client.DeleteFirewallDevice) (thanks, Tarun, for pointing this out). So if we're going the creation route, we should also incorporate this (if possible).

The linodego doesn't yet have the ability to list firewalls of a nodebalancer. The update feature can be added as soon as the ability has been added (https://jira.linode.com/browse/TPT-2535)

luthermonson:
> > Turns out it should be possible to remove firewalls from nodebalancer by using [DeleteFirewallDevice](https://pkg.go.dev/github.com/linode/linodego#Client.DeleteFirewallDevice) (thanks, Tarun, for pointing this out). So if we're going the creation route, we should also incorporate this (if possible).
> 
> The linodego doesn't yet have the ability to list firewalls of a nodebalancer. The update feature can be added as soon as the ability has been added (https://jira.linode.com/browse/TPT-2535)

https://github.com/linode/linode-cloud-controller-manager/issues/146

cloud/linode/loadbalancers_test.go
```diff
@@ -285,6 +290,37 @@ func testCreateNodeBalancer(t *testing.T, client *linodego.Client, _ *fakeAPI) {
 	}
 
 	defer func() { _ = lb.EnsureLoadBalancerDeleted(context.TODO(), "linodelb", svc) }()
+	return nil
+}
+
+func testCreateNodeBalancerWithOutFirewall(t *testing.T, client *linodego.Client, f *fakeAPI) {
+	err := testCreateNodeBalancer(t, client, f, nil)
+	if !reflect.DeepEqual(err, nil) {
```
okokes-akamai, line 298: No need to `DeepEqual` with nil, a simple comparison is sufficient. (same in the test below)

cloud/linode/loadbalancers_test.go
```diff
@@ -285,6 +290,37 @@ func testCreateNodeBalancer(t *testing.T, client *linodego.Client, _ *fakeAPI) {
...
+		t.Error("unexpected error")
```
okokes-akamai, line 299: Can you maybe squash these three lines (here and in the other two test cases) into a single line, something like `t.Fatalf("expected a nil error, got %v" err)`?

cloud/linode/loadbalancers_test.go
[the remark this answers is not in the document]
okokes-akamai: Looks good!

okokes-akamai (approved):

PR #147, opened 2023-12-01
Update Release Process

luthermonson:
* migrate to main branch
* redo gh workflows to use latest docker actions
* add helm chart to release artifact 
* redo dockerfile to look more like CSI and drop using `make build-linux`
* dropped unused linters
* dropped rand.Seed for go1.20
* drop io/ioutil usage


Dockerfile
```diff
@@ -1,5 +1,19 @@
-# NB: We now cross-compile the go binary locally using the Makefile
-FROM alpine:latest
+FROM golang:1.20-alpine as builder
```
okokes-akamai, line 1: We run CI on `stable` and `oldstable`, which now map to 1.21 and 1.20, correct? This means that these will point to 1.22 and 1.21 come Feb? I'm just worried that this 1.20 ref will become stale.

So my suggestions: 1. Bump this to 1.21 since it's the most up to date, 2. Move away from `stable`/`oldstable`.

Dockerfile
```diff
@@ -1,5 +1,19 @@
...
+COPY go.sum .
+COPY main.go .
+COPY cloud ./cloud
+COPY sentry ./sentry
+
+RUN go mod download
+RUN go build -a -ldflags '-extldflags "-static"' -o /bin/linode-cloud-controller-manager-linux /linode
+
+FROM alpine:3.18.4
 RUN apk update && apk add ca-certificates && rm -rf /var/cache/apk/*
```
okokes-akamai, line 15: `--no-cache` instead of `rm -rf`? Like `apk add -U --no-cache ca-certificates`

Dockerfile
luthermonson: i vote bumping to 1.21 in the dockerfile, the stable/oldstable thing is from dx and they prefer that after being burned by gh actions in the past.

okokes-akamai: Burned in what way? I'm worried _we'll_ be burned here as we run the risk of CI running on a different platform than our release pipeline.

luthermonson: @lgarber-akamai can elaborate but it sounded to me like doing stable/oldstable helped dx find go problems faster during the PR process instead of waiting last minute. i guess it's not being "burned by gh actions" so much as gh inaction?

if i had to choose my personal preference is the dockerfile is pinned to the go.mod version and the `make build` job has stable/oldstable... this pins the release artifact to the project's go version and allows the PR process to always have latest go to find problems faster.

schinmai-akamai (approved):

luthermonson:
@okokes-akamai

okokes-akamai:
@luthermonson I don't like the stable/oldstable stuff as it's completely divorced from whatever Go version we're actually using to build and run our code. Waiting for that thread to be resolved.

Dockerfile
lgarber-akamai: Sorry for the late follow-up on this 🙂 

Looking back it'd probably make sense to explicitly pin the CI, Dockerfile, and go.mod to go 1.21 (or 1.20) to keep things consistent.

okokes-akamai (approved):

luthermonson:
@okokes-akamai added an additional commit to drop `io/ioutil` to pass linting after your merges

Issue #144, opened 2023-12-01
Upgrade golangci-lint to 1.55.2


luthermonson:
As a cleanup task we need to upgrade to the latest linter and fix all linting errors in the process.

PR #148, opened 2023-12-02
upgrading golangci-lint to v1.55.2

luthermonson:
* set unused func params to _
* rewrite ok logic to remove superfluous else

### DIFF
--- a/cloud/linode/loadbalancers.go
+++ b/cloud/linode/loadbalancers.go
@@ -187,7 +187,7 @@ func (l *loadbalancers) cleanupOldNodeBalancer(ctx context.Context, service *v1.
 // GetLoadBalancerName returns the name of the load balancer.
 //
 // GetLoadBalancer will not modify service.
-func (l *loadbalancers) GetLoadBalancerName(ctx context.Context, clusterName string, service *v1.Service) string {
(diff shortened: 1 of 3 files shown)


luthermonson:
fixes #144

cloud/linode/loadbalancers.go
```diff
@@ -187,7 +187,7 @@ func (l *loadbalancers) cleanupOldNodeBalancer(ctx context.Context, service *v1.
 // GetLoadBalancerName returns the name of the load balancer.
 //
 // GetLoadBalancer will not modify service.
-func (l *loadbalancers) GetLoadBalancerName(ctx context.Context, clusterName string, service *v1.Service) string {
+func (l *loadbalancers) GetLoadBalancerName(_ context.Context, _ string, _ *v1.Service) string {
 	unixNano := strconv.FormatInt(time.Now().UnixNano(), 16)
```
schinmai-akamai, line 191: I think we should start adding in the clusterName here to the nb name. thoughts?

schinmai-akamai (approved):

cloud/linode/loadbalancers.go
luthermonson: can discuss further https://github.com/linode/linode-cloud-controller-manager/issues/150

merging this PR since it's kinda outta conext for fixing the linter

PR #162, opened 2024-01-04
feat: add linode UUID as a metadata to each node

shanduur:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

---

A host UUID was added to the Linode Instance in a client change: https://github.com/linode/linodego/pull/284

Apply this UUID from the CCM to expose the underlying host uniqueness, can improve scheduling and placement decisions for resources deployed on clusters with CCM installed.

### DIFF
--- a/cloud/linode/annotations.go
+++ b/cloud/linode/annotations.go
@@ -1,0 +1,33 @@
+package linode
+
+const (
+	// annLinodeDefaultProtocol is the annotation used to specify the default protocol
+	// for Linode load balancers. Options are tcp, http and https. Defaults to tcp.
+	annLinodeDefaultProtocol      = "service.beta.kubernetes.io/linode-loadbalancer-default-protocol"
+	annLinodePortConfigPrefix     = "service.beta.kubernetes.io/linode-loadbalancer-port-"
+	annLinodeDefaultProxyProtocol = "service.beta.kubernetes.io/linode-loadbalancer-default-proxy-protocol"
+
+	annLinodeCheckPath       = "service.beta.kubernetes.io/linode-loadbalancer-check-path"
+	annLinodeCheckBody       = "service.beta.kubernetes.io/linode-loadbalancer-check-body"
+	annLinodeHealthCheckType = "service.beta.kubernetes.io/linode-loadbalancer-check-type"
+
+	annLinodeHealthCheckInterval = "service.beta.kubernetes.io/linode-loadbalancer-check-interval"
+	annLinodeHealthCheckTimeout  = "service.beta.kubernetes.io/linode-loadbalancer-check-timeout"
+	annLinodeHealthCheckAttempts = "service.beta.kubernetes.io/linode-loadbalancer-check-attempts"
+	annLinodeHealthCheckPassive  = "service.beta.kubernetes.io/linode-loadbalancer-check-passive"
+
+	// annLinodeThrottle is the annotation specifying the value of the Client Connection
+	// Throttle, which limits the number of subsequent new connections per second from the
+	// same client IP. Options are a number between 1-20, or 0 to disable. Defaults to 20.
+	annLinodeThrottle = "service.beta.kubernetes.io/linode-loadbalancer-throttle"
+
+	annLinodeLoadBalancerPreserve = "service.beta.kubernetes.io/linode-loadbalancer-preserve"
+	annLinodeNodeBalancerID       = "service.beta.kubernetes.io/linode-loadbalancer-nodebalancer-id"
+
+	annLinodeHostnameOnlyIngress = "service.beta.kubernetes.io/linode-loadbalancer-hostname-only-ingress"
+	annLinodeLoadBalancerTags    = "service.beta.kubernetes.io/linode-loadbalancer-tags"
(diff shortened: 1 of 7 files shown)


cloud/linode/annotations.go
```diff
@@ -0,0 +1,33 @@
...
+
+	annLinodeLoadBalancerPreserve = "service.beta.kubernetes.io/linode-loadbalancer-preserve"
+	annLinodeNodeBalancerID       = "service.beta.kubernetes.io/linode-loadbalancer-nodebalancer-id"
+
+	annLinodeHostnameOnlyIngress = "service.beta.kubernetes.io/linode-loadbalancer-hostname-only-ingress"
+	annLinodeLoadBalancerTags    = "service.beta.kubernetes.io/linode-loadbalancer-tags"
+	annLinodeCloudFirewallID     = "service.beta.kubernetes.io/linode-loadbalancer-firewall-id"
+
+	annLinodeNodePrivateIP = "node.k8s.linode.com/private-ip"
+	annLinodeUUID          = "node.k8s.linode.com/uuid"
```
schinmai-akamai, line 32: I think this should say `host-uuid`

cloud/linode/client.go
```diff
@@ -27,3 +29,33 @@ type Client interface {
...
+
+	validatedURL := &url.URL{
+		Host:   parsedURL.Host,
+		Scheme: parsedURL.Scheme,
+	}
+
+	linodeClient.SetBaseURL(validatedURL.String())
+
+	version := ""
+	matches := regexp.MustCompile(`/v\d+`).FindAllString(parsedURL.Path, -1)
```
schinmai-akamai, line 52: is this still required after the linodego update?

cloud/linode/node_controller.go
```diff
@@ -0,0 +1,124 @@
...
+			klog.Infof("NodeController will handle updated node (%s) metadata", node.Name)
+			s.queue.Add(node)
+		},
+	})
+
+	go wait.Until(s.worker, time.Second, stopCh)
+	s.informer.Informer().Run(stopCh)
+}
+
+// worker runs a worker thread that dequeues deleted services and processes
```
schinmai-akamai, line 64: comment needs an update

srust (changes requested):
looks good - agree with `uuid` -> `host-uuid` change.

cloud/linode/client.go
shanduur: This is the same logic as in the [linode-cosi-driver](https://github.com/linode/linode-cosi-driver/blob/main/pkg/linodeclient/linodeclient.go#L43-L76) and [linode-blockstorage-csi-driver](https://github.com/linode/linode-blockstorage-csi-driver/blob/main/pkg/linode-client/linode-client.go#L34-L75), added there just so we have similar code paths in all 3 components.

cloud/linode/annotations.go
```diff
@@ -0,0 +1,33 @@
...
+
+	annLinodeLoadBalancerPreserve = "service.beta.kubernetes.io/linode-loadbalancer-preserve"
+	annLinodeNodeBalancerID       = "service.beta.kubernetes.io/linode-loadbalancer-nodebalancer-id"
+
+	annLinodeHostnameOnlyIngress = "service.beta.kubernetes.io/linode-loadbalancer-hostname-only-ingress"
+	annLinodeLoadBalancerTags    = "service.beta.kubernetes.io/linode-loadbalancer-tags"
+	annLinodeCloudFirewallID     = "service.beta.kubernetes.io/linode-loadbalancer-firewall-id"
+
+	annLinodeNodePrivateIP = "node.k8s.linode.com/private-ip"
+	annLinodeUUID          = "node.k8s.linode.com/host-uuid"
```
srust, line 32: sorry to be a pain - can the `annLinodeUUID` also be `annLinodeHostUUID` here. Someone in the future may glance at this quickly and think it corresponds to the VM.

cloud/linode/client.go
luthermonson: maybe a dumb question, but why dont we make a `linodeClient.ParseURL` which does this logic and sets the base url and version for you. this would move all this logic to the linodego and we could reuse it right?

luthermonson: @lgarber-akamai @zliang-akamai for visibility on my dumb questions

srust (approved):

cloud/linode/client.go
zliang-akamai: Hi @luthermonson, I think you are right. Let us discuss it and get back to you next week.

cloud/linode/client.go
```diff
@@ -27,3 +29,33 @@ type Client interface {
...
+
+	if len(matches) > 0 {
+		version = matches[len(matches)-1]
```
zliang-akamai, line 55:
```suggestion
		version = strings.Trim(matches[len(matches)-1], "/")
```

Current version of linodogo would escape `/v4` to be `%2Fv4` internally.

zliang-akamai (approved):

cloud/linode/client.go
lgarber-akamai: Sorry for the late follow-up, I think this is a good idea ^

PR #163, opened 2024-01-08
getting e2e tests working against LKE

luthermonson:
Gets e2e semi functional against an LKE cluster so we can make a release, this gets 19/20 tests working, the add/delete node test uses the terraform bits we will eventually remove and the test will get rewritten when that work is done.

Brief synopsis, this PR does...
* upgrades ginkgo
* adds timeout and api url as test args
* drops glog
* upgrades core kube components to 0.23.17 for the testing suite, this update is coming to ccm soon

To get tests running against an LKE cluster do the following:
1. boot an LKE cluster in your preferred region and download the kubeconfig
2. build your ccm locally and push to a public docker repo
3. install the ccm using the helm chart and your image or a [manifest](https://gist.github.com/luthermonson/361c4310ef234cba61202411b1c56bff)
4. export your SUITE_ARGS and LINODE_TOKEN, my args were `export SUITE_ARGS="--region=us-sea --use-existing --timeout=5m --kubeconfig=/Users/lmonson/l/linode-cloud-controller-manager/e2e/test-cluster-kubeconfig.yaml --linode-url https://api.linode.com/"`

This gets 18/20 tests to pass and you get something like the following output...

```
Summarizing 2 Failures:
  [FAIL] e2e tests Test LoadBalancer With HTTP updating to have HTTPS [BeforeEach] should reach the pods
  /Users/lmonson/l/linode-cloud-controller-manager/e2e/test/ccm_e2e_test.go:90
  [FAIL] e2e tests Test LoadBalancer With Node Addition [It] should reach the same pod every time it requests
  /Users/lmonson/l/linode-cloud-controller-manager/e2e/test/ccm_e2e_test.go:131

Ran 20 of 20 Specs in 722.276 seconds
FAIL! -- 18 Passed | 2 Failed | 0 Pending | 0 Skipped
--- FAIL: TestE2e (722.28s)
FAIL

Ginkgo ran 1 suite in 12m6.156576667s
```

When you use TEST_ARGS and run only the http/s test using `'--focus="With HTTP updating to have HTTPS"'` you can then get that test to pass when run standalone.

The final test `With Node Addition` uses the terraform to add/remove a node and we will be rewritten.

TODO in another PR
1. remove the terraform and rewrite the add/remove node test to use LKE api
2. update gh actions to create an lke cluster to use for tests
3. tests need to cleanup better of nodebalancers


schinmai-akamai (approved):

okokes-akamai:
> When you use TEST_ARGS and run only the http/s test using '--focus="With HTTP updating to have HTTPS"' you can then get that test to pass when run standalone.

If the test succeeds as standalone, it should be marked `Serial` and the ginkgo runner will ensure it runs without any other tests running at the same time.

e2e/test/ccm_e2e_test.go
```diff
@@ -231,11 +231,17 @@ var _ = Describe("e2e tests", func() {
 	}
 
 	addNewNode := func() {
+		if useExisting {
```
okokes-akamai, line 234: This doesn't look like something we want to have in our codebase? Maybe just conditionally skip the test instead?

luthermonson: ya i can revert this and do a bigger PR where we drop all the tf stuff

luthermonson:
> If the test succeeds as standalone, it should be marked `Serial` and the ginkgo runner will ensure it runs without any other tests running at the same time.

```
Ran 19 of 20 Specs in 500.232 seconds
SUCCESS! -- 19 Passed | 0 Failed | 0 Pending | 1 Skipped
PASS
```
You are the ginkgo zenmaster 🙇 all your changes are accounted for and they run unassisted. I'm going to merge to try and still get the release done tonight for @mabojars

PR #167, opened 2024-01-12
fix: set host-uuid correctly on non LKE clusters

schinmai-akamai:
Fixes a bug where the host-uuid was not set correctly for non-LKE clusters. 

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/cloud/linode/instances_test.go
+++ b/cloud/linode/instances_test.go
@@ -233,17 +233,6 @@ func TestMalformedProviders(t *testing.T) {
 
 	client := NewMockClient(ctrl)
 
-	t.Run("fails on malformed providerID", func(t *testing.T) {
-		instances := newInstances(client)
-		providerID := "bogus://bogus"
-		node := nodeWithProviderID(providerID)
-		client.EXPECT().ListInstances(gomock.Any(), nil).Times(1).Return([]linodego.Instance{}, nil)
-
-		meta, err := instances.InstanceMetadata(ctx, node)
-		assert.ErrorIs(t, err, invalidProviderIDError{providerID})
-		assert.Nil(t, meta)
-	})
-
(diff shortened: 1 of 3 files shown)


srust (approved):
This begs the question - if it works with or without a linode provider ID, what is the linode provider ID useful for.

It appears the answer is sentry only. Is that your understanding as well?

schinmai-akamai:
> This begs the question - if it works with or without a linode provider ID, what is the linode provider ID useful for.
> 
> It appears the answer is sentry only. Is that your understanding as well?

The lookup using providerID is much faster than the lookup by name - see instances.go `linodeByName` - it does a O(n) search among all available nodes, while `linodeByID` does a O(1) map read. 

I left the code as is, to take the efficiency hit only in non-LKE clusters

cloud/linode/instances_test.go
```diff
@@ -233,17 +233,6 @@ func TestMalformedProviders(t *testing.T) {
 
 	client := NewMockClient(ctrl)
 
-	t.Run("fails on malformed providerID", func(t *testing.T) {
```
luthermonson, line 236: why remove? can't fix?

schinmai-akamai: there's no "malformed provider ID" anymore, its either a linode-id or a label that we can look up - since we mock the linodeGo response, I didn't feel like it was a useful test to see if a string existed in a list (that we mock)

okokes-akamai: This is not what the test does, it doesn't test anything in the linodego response - it checks that we error out in case there's an incorrect `spec` value on the `Node` object. The new codebase just ignores this issue entirely and I don't find that to be correct behaviour.

PR #168, opened 2024-01-14
always return lbnotfound from getNodeBalancerForService

luthermonson:
Found an edge case where I had deleted a nodebalancer but it was set on a service and wouldn't clear the finalizer and never saw the short-circuit log from here: https://github.com/linode/linode-cloud-controller-manager/blob/main/cloud/linode/loadbalancers.go#L413

the lbnotfound type check was rewriting to a new error and never returning so moving the onus on checking lbnotfound to the create path so a second load balancer is not created and the skip now works.

### DIFF
--- a/e2e/test/ccm_e2e_test.go
+++ b/e2e/test/ccm_e2e_test.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"os/exec"
 	"strconv"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/linode/linodego"
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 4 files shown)


okokes-akamai:
Can you reflect this change in tests? Fix the failing one and perhaps add some new ones.

luthermonson:
> Can you reflect this change in tests? Fix the failing one and perhaps add some new ones.

tests added, fail in main but pass in my branch

e2e/test/ccm_e2e_test.go
```diff
@@ -90,6 +92,14 @@ var _ = Describe("e2e tests", func() {
 		Eventually(watcher.ResultChan()).Should(Receive(EnsuredService()))
 	}
 
+	ensureServiceWasDeleted := func() {
+		notfound := func() bool {
+			_, err := f.LoadBalancer.GetService()
+			return errors.IsNotFound(err)
+		}
+		Eventually(notfound).WithTimeout(10 * time.Second).Should(BeTrue())
```
okokes-akamai, line 100: This is impossible to debug, please return the error and check that it's not found, so that we know - in case of failure - if it was nil or a different error than not found.

luthermonson: great feedback, this has been refactored

okokes-akamai (approved):

schinmai-akamai (approved):

PR #169, opened 2024-01-18
CCM Managed firewalls

### LINKED ISSUE #152: Add Firewall ACL Rules as an Annotation
Now that we have firewall support we should add to the story to passing in information for to create the firewall and access rules. Add enough annotations to create a firewall with ACLs, create it and add it to the nodebalancer. Bonus points... due to account limits try to find out a way to reuse firewalls if the ACL rules are identical. This might be easy with config maps? Perhaps the rules are in there and when to service annotations match the same config map you only use the one resource in Linode.

schinmai-akamai:
Adds a new feature - ccm managed firewalls. Firewall rules can by specified by the user using an annotation - `service.beta.kubernetes.io/linode-loadbalancer-firewall-acl` 
closes #152 

### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 


cloud/linode/loadbalancers.go
```diff
@@ -246,55 +249,36 @@ func (l *loadbalancers) getNodeBalancerDeviceId(ctx context.Context, firewallID,
...
 }
 
-// updateNodeBalancerFirewall updates the firewall attached to the nodebalancer
-//
-// Firewall is updated in 3 scenario's:
-// - Add new firewall to the nodeBalancer
-// - Update firewall attached to the nodebalancer
-// - Remove firewall attached
-func (l *loadbalancers) updateNodeBalancerFirewall(ctx context.Context, service *v1.Service, nb *linodego.NodeBalancer) error {
+func (l *loadbalancers) updateFWwithID(ctx context.Context, service *v1.Service, nb *linodego.NodeBalancer) error {
```
okokes-akamai, line 252: Please add docs here + maybe expand the name, no need to shorten firewall to FW.

cloud/linode/loadbalancers.go
```diff
@@ -246,55 +249,36 @@ func (l *loadbalancers) getNodeBalancerDeviceId(ctx context.Context, firewallID,
...
-	// get the ID of the firewall that is already attached to the nodeBalancer.
+	// get the ID of the firewall that is already attached to the nodeBalancer, if we have one.
 	var existingFirewallID int
-	if len(firewalls) != 0 {
+	if len(firewalls) == 1 {
 		existingFirewallID = firewalls[0].ID
 	}
 
-	// if existing firewall and new firewall differs, update accordingly.
+	// if existing firewall and new firewall differs,
```
okokes-akamai, line 277: trailing comma... maybe just write out what happens now, otherwise the comment is not super helpful

cloud/linode/loadbalancers.go
```diff
@@ -309,16 +293,120 @@ func (l *loadbalancers) updateNodeBalancerFirewall(ctx context.Context, service
...
 				Type: "nodebalancer",
 			})
 			if err != nil {
 				return err
 			}
 		}
+	case 1:
+		{
+			// We do not want to get into the complexity of reconciling differences, might as well just pull what's in the svc annotation now and update the fw.
+			fwCreateOpts, err := l.createFirewallOptsForSvc(service.Name, []string{""}, service)
```
okokes-akamai, line 340: Are we worried about API chatter? I understand diffing these might be laborious, but could be worth it in terms of the number of mutating calls.

cloud/linode/loadbalancers.go
```diff
@@ -309,16 +293,120 @@ func (l *loadbalancers) updateNodeBalancerFirewall(ctx context.Context, service
...
+	}
+	return nil
+}
+
+// updateNodeBalancerFirewall reconciles the firewall attached to the nodebalancer
+//
+// This function does the follwing
+//  1. If a firewallID annotation is present, it checks if the nodebalancer has a firewall attached, and if it matches the annotationID
+//     a. If the IDs match, nothing to do here.
+//     b. If they don't match, then the nb is removed from the old FW, and its attached to the new firewall
```
okokes-akamai, line 360: Maybe irrelevant, but is this a safe ordering of steps? Wouldn't we want to first attach a new FW and then detach the old one?

Because should anything happen in the interim, the NB will be left unprotected.

cloud/linode/loadbalancers.go
```diff
@@ -309,16 +293,120 @@ func (l *loadbalancers) updateNodeBalancerFirewall(ctx context.Context, service
...
+//  3. If neither of these annotations are present,
+//	  a. AND if no firewalls are attached to the nodebalancer, nothing to do.
+//	  b. if the NB has ONE firewall attached, remove it from nb, and clean up if nothing else is attached to it
+//	  c. If there are more than one fw attached to it, then its a problem, return an err
+// IF a user creates a fw ID externally, and then switches to using a ACL, the CCM will take over the fw that's attached to the nodebalancer.
+
+func (l *loadbalancers) updateNodeBalancerFirewall(ctx context.Context, service *v1.Service, nb *linodego.NodeBalancer) error {
+	// get the new firewall id from the annotation (if any).
+	_, fwIDExists := service.GetAnnotations()[annLinodeCloudFirewallID]
+	if fwIDExists { // If an ID exists, we ignore everything else and handle just that
```
okokes-akamai, line 373: > we ignore everything else and handle just that

What about the case when _both_ annotations are present?

cloud/linode/loadbalancers.go
```diff
@@ -309,16 +293,120 @@ func (l *loadbalancers) updateNodeBalancerFirewall(ctx context.Context, service
...
+	if fwACLExists { // if an ACL exists, but no ID, just update the ACL on the fw.
+		return l.updateFWwithACL(ctx, service, nb)
+	}
+
+	// No firewall ID or ACL annotation, see if there are firewalls attached to our nb
+	firewalls, err := l.client.ListNodeBalancerFirewalls(ctx, nb.ID, &linodego.ListOptions{})
+	if err != nil {
+		return err
+	}
+	switch len(firewalls) {
```
okokes-akamai, line 388: nit: in the interest of keeping code to the left, I'd maybe early exit in the 0/2+ cases and then continue execution in the "happy" path of 1.

```go
if len(firewalls) == 0 {
  return nil
}
if len(firewalls) > 1 {
  return errTooManyFirewalls
}

err := l.client.Delete...
```

cloud/linode/loadbalancers.go
```diff
@@ -572,14 +660,77 @@ func (l *loadbalancers) getNodeBalancerByID(ctx context.Context, service *v1.Ser
...
+	}
+
 	tagStr, ok := service.GetAnnotations()[annLinodeLoadBalancerTags]
 	if ok {
 		return append(tags, strings.Split(tagStr, ",")...)
 	}
 	return tags
 }
 
+func processACL(fwcreateOpts *linodego.FirewallCreateOptions, aclType, label, svcName, ports string, ips linodego.NetworkAddresses) {
```
okokes-akamai, line 675: this doesn't have docs + it's not quite clear from the name (or quite frankly from reading the code) what this is supposed to do (it only started becoming somewhat clear from context)

cloud/linode/loadbalancers.go
```diff
@@ -572,14 +660,77 @@ func (l *loadbalancers) getNodeBalancerByID(ctx context.Context, service *v1.Ser
...
+	// Fetch acl from annotation
+	aclString := svc.GetAnnotations()[annLinodeCloudFirewallACL]
+	fwcreateOpts := linodego.FirewallCreateOptions{
+		Label: label,
+		Tags:  tags,
+	}
+
+	servicePorts := make([]string, len(svc.Spec.Ports))
+	for idx, port := range svc.Spec.Ports {
+		servicePorts[idx] = strconv.Itoa(int(port.Port))
```
okokes-akamai, line 709: Go antipattern, I'd rather `append`. The `servicePorts` definition will then have `len(svc.Spec.Ports)` as its capacity, not length.

cloud/linode/loadbalancers.go
schinmai-akamai: added a line to clarify that

cloud/linode/loadbalancers.go
schinmai-akamai: changed the order of operations

cloud/linode/loadbalancers.go
```diff
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"reflect"
+	"slices"
```
schinmai-akamai, line 11: @okokes-akamai i couldn't find a way to make this work on go 1.20 && 1.21 - should I default to using x/exp/slices until we get rid of 1.20 ?

cloud/linode/loadbalancers.go
okokes-akamai: It wasn't happy about it, because you didn't do what I suggested :-)

You didn't initialise the slice (unlike now).

```
servicePorts := make([]string, 0, len(svc.Spec.Ports))
```

image

cloud/linode/loadbalancers.go
okokes-akamai: Well, we already build CCM with 1.21, so I'm not sure why we're testing this against 1.20. I wasn't happy about this back when I reviewed this PR: https://github.com/linode/linode-cloud-controller-manager/pull/147#discussion_r1413611112

Also, in about three weeks, `oldstable` will start pointing to 1.21, so we'll get rid of this automatically, but I'd rather drop the stable/oldstable altogether.

schinmai-akamai: will create a follow up oldstable stuff - keeping `x/exp/slices` for now

cloud/linode/loadbalancers.go
```diff
@@ -24,7 +26,11 @@ import (
 	"github.com/linode/linodego"
 )
 
-var errNoNodesAvailable = errors.New("no nodes available for nodebalancer")
+var (
+	errNoNodesAvailable = errors.New("no nodes available for nodebalancer")
```
luthermonson, line 30: capitalize `No` to match the others?

cloud/linode/loadbalancers.go
```diff
@@ -24,7 +26,11 @@ import (
...
+	errInvalidFWConfig  = errors.New("Specify either an allowList or a denyList for a firewall")
+	errTooManyFirewalls = errors.New("Too  many firewalls attached to a nodebalancer")
```
luthermonson, line 32:
```suggestion
	errTooManyFirewalls = errors.New("Too many firewalls attached to a nodebalancer")
```

cloud/linode/loadbalancers.go
luthermonson, line 30:
```suggestion
	errNoNodesAvailable = errors.New("No nodes available for nodebalancer")
```

cloud/linode/loadbalancers.go
```diff
@@ -227,8 +232,8 @@ func (l *loadbalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
 }
 
 // getNodeBalancerDeviceId gets the deviceID of the nodeBalancer that is attached to the firewall.
-func (l *loadbalancers) getNodeBalancerDeviceId(ctx context.Context, firewallID, nbID, page int) (int, bool, error) {
-	devices, err := l.client.ListFirewallDevices(ctx, firewallID, &linodego.ListOptions{PageSize: 500, PageOptions: &linodego.PageOptions{Page: page}})
+func (l *loadbalancers) getNodeBalancerDeviceId(ctx context.Context, firewallID, nbID int) (int, bool, error) {
```
luthermonson, line 235:
```suggestion
func (l *loadbalancers) getNodeBalancerDeviceID(ctx context.Context, firewallID, nbID int) (int, bool, error) {
```

cloud/linode/loadbalancers.go
```diff
@@ -24,7 +26,11 @@ import (
 	"github.com/linode/linodego"
 )
 
-var errNoNodesAvailable = errors.New("no nodes available for nodebalancer")
+var (
+	errNoNodesAvailable = errors.New("No nodes available for nodebalancer")
+	errInvalidFWConfig  = errors.New("Specify either an allowList or a denyList for a firewall")
+	errTooManyFirewalls = errors.New("Too many firewalls attached to a nodebalancer")
```
luthermonson, line 32: how do we even get to the too many firewalls error, just if a user attached a firewall to the nodebalancer by hand? should we offer additional error messages for which firewalls we found so they can go clean up themselves if this happens?

schinmai-akamai: Yes, we'd get there if the user attaches multiple firewalls by hand - I wanted to be defensive here and not remove stuff the ccm didn't create. I will add a line printing the IDs so the user can investigate

okokes-akamai (approved):

PR #176, opened 2024-02-12
update default CCM node selector and tolerations from master to control-plane

AshleyDumaine:
The selector for the CCM should be updated to account for the [deprecation and removal](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-master-taint) of the `master` node-role. Currently, this doesn't work out-of-the-box on k8s 1.29 which uses only `node-role.kubernetes.io/control-plane` for control plane nodes.


rahulait:
The change looks good to me.

avestuk (approved):

eljohnson92 (approved):

tchinmai7 (approved):

PR #141, opened 2023-11-13
Allow specifying Private IP by annotation for VLAN / VPC support

glennpratt:
Fixes #111

Kubernetes node internal IP addresses are not necessarily Linode Private IPs, which must be used for NodeBalancers to communicate with a node. This allows specifying the IP with an annotation in such cases.

### General:

* [x] Have you removed all sensitive information, including but not limited to access keys and passwords?
* [x] Have you checked to ensure there aren't other open or closed [Pull Requests](../../pulls) for the same bug/feature/question?

### Pull Request Guidelines:

1. [x] Does your submission pass tests?
1. [x] Have you added tests? 
1. [x] Are you addressing a single feature in this PR? 
1. [x] Are your commits atomic, addressing one change per commit?
1. [x] Are you following the conventions of the language? 
1. [x] Have you saved your large formatting changes for a different PR, so we can focus on your work?
1. [x] Have you explained your rationale for why this feature is needed? 
1. [x] Have you linked your PR to an [open issue](https://blog.github.com/2013-05-14-closing-issues-via-pull-requests/)

### Testing

Manually set the annotation to on dev2-worker-0 to another Linode (that isn't running these services) to prove the change: 

image

The canonical test would be on a VLAN/VPC cluster, but this seem sufficient.

PR #182, opened 2024-02-21
Auto add private ip annotation to nodes

rahulait:
We added support for specifying Private IP by annotation for VLAN / VPC in PR https://github.com/linode/linode-cloud-controller-manager/pull/141. We wanted to further enhance it and automatically add annotations to nodes with the backend private ip address. This PR enables automatically adding the annotation by looking at node's private ip-address.

#### Key points
Currently, linode nodebalancer doesn't support any subnet outside of 192.168.128.0/17. Once nodebalancer adds support for other subnets, we need to update linode-ccm to support that.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/cloud/linode/node_controller.go
+++ b/cloud/linode/node_controller.go
@@ -2,6 +2,8 @@ package linode
 
 import (
 	"context"
+	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"strconv"
@@ -127,8 +129,14 @@ func (s *nodeController) handleNode(ctx context.Context, node *v1.Node) error {
 
 	lastUpdate := s.LastMetadataUpdate(node.Name)
 
-	uuid, ok := node.Labels[annotations.AnnLinodeHostUUID]
-	if ok && time.Since(lastUpdate) < s.ttl {
+	uuid, foundLabel := node.Labels[annotations.AnnLinodeHostUUID]
+	configuredPrivateIP, foundAnnotation := node.Annotations[annotations.AnnLinodeNodePrivateIP]
+	if foundLabel && foundAnnotation && time.Since(lastUpdate) < s.ttl {
+		return nil
+	}
+
+	// if autoannotate is set to false, just check label and ttl
+	if foundLabel && !Options.AutoAnnotateNode && time.Since(lastUpdate) < s.ttl {
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -290,6 +290,15 @@ helm upgrade ccm-linode --install --set apiToken=$LINODE_API_TOKEN,region=$REGIO
 ```
 _See [helm upgrade](https://helm.sh/docs/helm/helm_upgrade/) for command documentation._
 
+### Flags
+If required, ccm-linode can be started with following linode specific parameters:
+
+Key | Values | Default | Description
+---|---|---|---
+`linodego-debug` | `boolean` | `false` | enables debug output for the LinodeAPI wrapper.
+`linode-node-private-subnet` | `net.IPNet` | `192.168.128.0/17` | specifies backend network to use with nodebalancers.
```
eljohnson92, line 299: could we potentially rename this to something like `linode-nodebalancer-subnet`?

cloud/linode/node_controller.go
```diff
@@ -129,13 +129,29 @@ func (s *nodeController) handleNode(ctx context.Context, node *v1.Node) error {
...
+	// TODO: Once nodebalancer supports other subnets, instead of using linode.IPv4
+	//       which contains public and private ip, use different call which returns
+	//       all ip's configured on the node
+	linodePrivIP := ""
+	for _, ip := range linode.IPv4 {
+		if Options.LinodeNodePrivateSubnet.Contains(*ip) {
+			linodePrivIP = ip.String()
+			break
+		}
+	}
```
glennpratt, line 152: The Linode Go client already has a method for getting and splitting these values:

https://github.com/linode/linodego/blob/852aace73a7c6b439176e9b361a607296589f37c/instance_ips.go#L77

https://github.com/linode/linodego/blob/852aace73a7c6b439176e9b361a607296589f37c/instance_ips.go#L10-L36

It's what the terraform provider relies on as well:

https://github.com/linode/terraform-provider-linode/blob/a148b9c6e95d552e42f05497900b6fdd0981e61f/linode/instance/flatten.go#L37-L39

If it were me, I'd match what the terraform provider does to get "private_ip_address".

AshleyDumaine: +1, We also do something similar on CAPL: https://github.com/linode/cluster-api-provider-linode/blob/main/cloud/services/loadbalancers.go#L122-L155

glennpratt:
Overall, excited to have this automated.

I would wire things up using the `GetInstanceIPAddresses` method (and running it through the cache too), then using the pattern Terraform and CAPL use.
I'd also drop the configuration flag. This pattern is pretty heavily used already. When NodeBalancers change to support other backends, other IP ranges or IPv6, I think changing the code and expecting users to update is better than trying to guess what config flags they need in advance.

rahulait:
@glennpratt We now need this feature as we have already added support to use vpc specific ip's as internal ip's if the cluster is running within the VPC (to prevent cilium from caching incorrect ips). Without this PR, linode-ccm tries to use whatever internal ip we have for node with nodebalancers and keeps failing. We'll set autoAnnotateNode to true when installing clusters within VPC and it will detect an ip on default private subnet ip address (192.168.128.0/17), annotate the node using it and then CCM can configure nodebalancers successfully.

rahulait:
@glennpratt Removed the flags and now relying on the linode API response for the private IP. PTAL 🙏

glennpratt (approved):

PR #195, opened 2024-03-21
revert PR 184, PR 190 and PR 182

rahulait:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


tchinmai7 (approved):
tested manually that it works - will wait for @luthermonson's confirmation

deploy/chart/templates/daemonset.yaml
```diff
@@ -33,18 +33,6 @@ spec:
             - --v=3
             - --port=0
             - --secure-port=10253
-            {{- if .Values.linodegoDebug }}
```
rosskirkpat, line 36: The `linodego-debug` flag is still present, we should ensure that the helm chart still supports enabling it post-revert.

rosskirkpat (changes requested):
The node cache TTL is configurable via an envvar (`LINODE_INSTANCE_CACHE_TTL`), however, the helm chart does not offer a method of setting that variable during deployment. 

I believe this should be added to allow fine tuning CCM performance as well as aiding in future debugging/troubleshooting.

rahulait:
> The node cache TTL is configurable via an envvar (`LINODE_INSTANCE_CACHE_TTL`), however, the helm chart does not offer a method of setting that variable during deployment.
> 
> I believe this should be added to allow fine tuning CCM performance as well as aiding in future debugging/troubleshooting.

I think we can do that by setting additional env vars here:
https://github.com/linode/linode-cloud-controller-manager/blob/main/deploy/chart/values.yaml#L53-L58

rosskirkpat (approved):

luthermonson (approved):

PR #196, opened 2024-03-26
Add route_controller to linode CCM

rahulait:
### General:
This PR reintroduces changes which were reverted in https://github.com/linode/linode-cloud-controller-manager/pull/196

Additional changes added to PR:
1. If we are running outside of VPC, we don't fetch InstanceConfigs
2. If we are running within VPC, we fetch InstanceConfigs for instances within the VPC

Note: In future, we would like to use /v4/vpcs/:id/ips to list all VPC ips so that we can get all instance's ips in just 2 calls than existing 1+len(nodes) calls.

Calls made every 20 secs in a 6 node cluster inside VPC (started CCM with linodego-debug flag set to true):
```
k logs --timestamps pod/ccm-linode-cbxwl -n kube-system -f | grep "/v4/"

2024-03-27T13:50:35.520144953Z GET  /v4/linode/instances  HTTP/1.1
2024-03-27T13:50:35.599625368Z GET  /v4/vpcs/41220  HTTP/1.1
2024-03-27T13:50:35.691136075Z GET  /v4/linode/instances/56323907/configs  HTTP/1.1
2024-03-27T13:50:35.696009496Z GET  /v4/linode/instances/56324011/configs  HTTP/1.1
2024-03-27T13:50:35.700781396Z GET  /v4/linode/instances/56323949/configs  HTTP/1.1
2024-03-27T13:50:35.701410791Z GET  /v4/linode/instances/56323970/configs  HTTP/1.1
2024-03-27T13:50:35.705268044Z GET  /v4/linode/instances/56323968/configs  HTTP/1.1
2024-03-27T13:50:35.727128327Z GET  /v4/linode/instances/56323974/configs  HTTP/1.1

2024-03-27T13:50:55.519257954Z GET  /v4/linode/instances  HTTP/1.1
2024-03-27T13:50:55.623009604Z GET  /v4/vpcs/41220  HTTP/1.1
2024-03-27T13:50:55.708633682Z GET  /v4/linode/instances/56324011/configs  HTTP/1.1
2024-03-27T13:50:55.721245148Z GET  /v4/linode/instances/56323970/configs  HTTP/1.1
2024-03-27T13:50:55.724830978Z GET  /v4/linode/instances/56323907/configs  HTTP/1.1
2024-03-27T13:50:55.737507344Z GET  /v4/linode/instances/56323974/configs  HTTP/1.1
2024-03-27T13:50:55.743463294Z GET  /v4/linode/instances/56323949/configs  HTTP/1.1
[...]

Issue #183, opened 2024-02-29
add route_controller to linode-cloud-controller-manager


rahulait:
## General:

* [x] Have you removed all sensitive information, including but not limited to access keys and passwords?
* [x] Have you checked to ensure there aren't other open or closed [Pull Requests](../../pulls) for the same bug/feature/question?

----

## Feature Requests:
* [x] Have you explained your rationale for why this feature is needed? 
* [x] Have you offered a proposed implementation/solution? 

Each node gets a podCIDR allocated by k8s. When k8s clusters are provisioned within VPCs, we would like to configure routes such that the pod traffic flows via the VPC interface. In short, we need to allow those subnets on the linodes in VPC. In cloud-controller-manager, we can leverage route_controller to add/update routes as nodes are added/updated.

We would like to add implementation for route_controller to linode-cloud-controller-manager so that it automatically updates routes when new podCIDRs are assigned to nodes.

rahulait:
Added the functionality in https://github.com/linode/linode-cloud-controller-manager/pull/199, closing this issue as done. This will be released in v0.4.4 version.

PR #199, opened 2024-04-10
feat: Add route_controller to linode CCM

rahulait:
### General:
Implements: https://github.com/linode/linode-cloud-controller-manager/issues/183
This PR reintroduces changes which were reverted in https://github.com/linode/linode-cloud-controller-manager/pull/196

Additional changes added to PR:
1. Instead of using InstanceConfigs, we use ListVPCIPAddresses
2. When running outside of VPC, we don't fetch VPC specific ip-addresses
2. If we are running within VPC, we use ListVPCIPAddresses to fetch VPC specific ips and ranges

Two calls made every 20 secs in a 6 node cluster inside VPC (started CCM with linodego-debug flag set to true).
```
k logs --timestamps pod/ccm-linode-cbxwl -n kube-system -f | grep "/v4/"

2024-04-16T21:37:19.131269749Z GET  /v4/linode/instances  HTTP/1.1
2024-04-16T21:37:19.209701813Z GET  /v4/vpcs/56102/ips?page=1  HTTP/1.1
2024-04-16T21:37:39.207576432Z GET  /v4/linode/instances  HTTP/1.1
2024-04-16T21:37:39.289152871Z GET  /v4/vpcs/56102/ips?page=1  HTTP/1.1
2024-04-16T21:37:59.153318333Z GET  /v4/linode/instances  HTTP/1.1
2024-04-16T21:37:59.230218661Z GET  /v4/vpcs/56102/ips?page=1  HTTP/1.1
2024-04-16T21:38:09.114433434Z GET  /v4/vpcs/56102/ips?page=1  HTTP/1.1
2024-04-16T21:38:19.141108393Z GET  /v4/linode/instances  HTTP/1.1
2024-04-16T21:38:19.221340056Z GET  /v4/vpcs/56102/ips?page=1  HTTP/1.1
```

When running outside of VPC, we don't have route-controller enabled and hence we don't get instance cache refreshes that often.
```
k logs --timestamps pod/ccm-linode-42bjf -n kube-system -f | grep "/v4"
2024-04-10T16:02:07.246780998Z GET  /v4/linode/instances  HTTP/1.1
2024-04-10T16:02:07.326557026Z GET  /v4/linode/instances  HTTP/1.1
2024-04-10T16:07:07.496047310Z GET  /v4/linode/instances  HTTP/1.1
```

Since route_controller gets triggered every 10 seconds and instance cache is valid for [15 seconds](https://github.com/linode/linode-cloud-controller-manager/blob/v0.4.2/cloud/linode/instances.go#L61-L66), each second iteration of route_controller (2*10) finds the instance cache expired and causes it to refresh. I would prefer to change route_controller's default triggering interval from 10 seconds to 60 seconds as its running too frequently and doing unnecessary work. Even if a node doesn't get its routes updated for 60 seconds when it joins the cluster, its fine and pods will come up fine, just that they won't be able to communicate with pods on other nodes for 60 seconds or so for the first time should be fine? I am open to discussions about it and come to some conclusion on what time we should set for route-reconciliation-period for route_controller.

**Update**: It was decided that making two calls every 20 seconds should not be a concern for API. It should be able to handle much larger load and us hitting two APIs every 20 seconds should not be a concern, so we are going to let route_controller run with its default triggering interval.

### Pull Request Guidelines:

1. [x] Have you added tests? 


rahulait:
@avestuk @okokes-akamai @tchinmai7 this PR is ready for review again. I have changed the API calls and now relying on 
/v4/vpcs/:id/ips to fetch ips within vpc and use them for route_controller. PTAL 🙏

cloud/linode/route_controller.go
```diff
@@ -0,0 +1,252 @@
...
+	intfVPCIP := linodego.VPCIP{}
+	for _, ir := range instanceRoutes {
+		if ir.Address != nil && ir.VPCID == r.vpcid {
+			intfVPCIP = ir
+			continue
+		}
+
+		if ir.Address != nil || ir.VPCID != r.vpcid {
+			continue
+		}
```
avestuk, line 146:
```suggestion
		if ir.VPCID != r.vpcid {
			continue
		}

		if ir.Address != nil {
			intfVPCIP = ir
			continue
		}
```

I think that having the logic laid out like this makes it a little easier to follow.

cloud/linode/route_controller.go
```diff
@@ -0,0 +1,252 @@
...
+func newRoutes(client client.Client) (cloudprovider.Routes, error) {
+	timeout := 60
+	if raw, ok := os.LookupEnv("LINODE_ROUTES_CACHE_TTL"); ok {
+		if t, _ := strconv.Atoi(raw); t > 0 {
+			timeout = t
+		}
+	}
+	klog.V(3).Infof("TTL for routeCache set to %d", timeout)
+
+	vpcid := vpcInfo.getID()
```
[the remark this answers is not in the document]
rahulait: We can. I thought in future other controllers like node_controller or service_controller etc might also want to use vpcid and then we would be passing the same variable everywhere. Since its kind of global value shared by all and set once when the program inits, I kept it as a global var. Let me know if you still think passing it would be better and I can add the change.

cloud/linode/cloud.go
```diff
@@ -25,14 +26,46 @@ const (
...
+	KubeconfigFlag        *pflag.Flag
+	LinodeGoDebug         bool
+	EnableRouteController bool
+	VPCName               string
 }
 
+// vpcDetails is set when VPCName options flag is set.
+// We use it to list instances running within the VPC if set
+type vpcDetails struct {
+	mu   sync.RWMutex
```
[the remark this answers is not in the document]
rahulait: This was just to keep code safe just in case some gotest when run in parallel invokes init in multiple goroutines. I agree it doesn't and will most likely never be, was one of the suggestions in earlier reviews just to make global variable thread safe and hence used mutex here as well.

cloud/linode/route_controller.go
```diff
@@ -0,0 +1,252 @@
...
+type routes struct {
+	vpcid      int
+	client     client.Client
+	instances  *instances
+	routeCache *routeCache
+}
+
+func newRoutes(client client.Client) (cloudprovider.Routes, error) {
+	timeout := 60
+	if raw, ok := os.LookupEnv("LINODE_ROUTES_CACHE_TTL"); ok {
```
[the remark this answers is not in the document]
rahulait: Thanks, updated

cloud/linode/route_controller.go
rahulait: Thanks, updated

deploy/chart/templates/daemonset.yaml
```diff
@@ -35,6 +35,15 @@ spec:
             {{- if .Values.linodegoDebug }}
             - --linodego-debug={{ .Values.linodegoDebug }}
             {{- end }}
+            {{- if .Values.routeController }}
+            - --enable-route-controller=true
+            - --vpc-name={{ .Values.routeController.vpcName }}
```
tchinmai7, line 40: is there a way for the ccm to figure this out automatically?

rahulait: Thats a good question. If we change the logic to always fetch instance ips and vpc ips, build a cache and check if instance is present in both, then we can say this particular instance is in VPC. However, when route_controller calls listRoutes, it will not know which instance is part of particular VPC and will list all routes configured on all instances in an account. This will result route_controller to send delete for routes of instances it doesn't expect in the cluster. It might be possible that we wait for createRoute call for an instance and then get its interfaces and that will tell us which vpc its part of, and then use the result going further, but I feel we might hit some edge cases with it. There are inbuilt flags in [cloud-provider](https://github.com/kubernetes/cloud-provider/blob/master/config/types.go#L71) but those haven't been useful, so I had to introduce this new one. Detecting it automatically would be so nice to have. Maybe something we can keep an eye on and if we find a way in future, switch to it.

cloud/linode/route_controller.go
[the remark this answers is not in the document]
rahulait: I am going to keep it as it is for now as in future, we might want to auto detect the VPC where CCM is running and in such cases, having a global variable and setting VPC id there would be useful for other controllers to consume once it is determined.

avestuk (approved):

rahulait:
Ran e2e tests with updated image and they passed:
```
[AfterSuite] PASSED [0.029 seconds]
------------------------------

Ran 22 of 23 Specs in 396.333 seconds
SUCCESS! -- 22 Passed | 0 Failed | 0 Pending | 1 Skipped
PASS

Ginkgo ran 1 suite in 6m39.109931352s
Test Suite Passed
```

deploy/chart/templates/daemonset.yaml
```diff
@@ -36,6 +36,15 @@ spec:
...
```
tchinmai7, line 41: nit: If vpc-name is required, we should verify that in the helm chart and error if its not provided

tchinmai7 (approved):

tchinmai7 (approved):

deploy/chart/templates/daemonset.yaml
rahulait: Updated

PR #208, opened 2024-05-03
[feat] - add support for cilium-backed loadbalancers

AshleyDumaine:
This PR is to support cases where the linode-CCM is installed but instead of backing a LoadBalancer Service via NodeBalancers, the user wants to leverage [Cilium's BGP load balancing](https://docs.cilium.io/en/stable/network/bgp-control-plane/#service-announcements) with automatically requested and assigned shared IPs.

### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 


cloud/linode/cilium_loadbalancers.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// To hold the IP in lieu of a proper IP reservation system, a special Nanode is
+// created but not booted and used to hold all shared IPs.
+func (l *loadbalancers) ensureIPHolder(ctx context.Context) (*linodego.Instance, error) {
+	ipHolder, err := l.getIPHolder(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if ipHolder == nil {
```
eljohnson92, line 136: can we change this to
```
if ipHolder != nil {
    return ipHolder, nil
}
...
```

cloud/linode/cilium_loadbalancers.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if ipHolder != nil {
+		return ipHolder, nil
+	}
+
+	ipHolder, err = l.client.CreateInstance(ctx, linodego.InstanceCreateOptions{
+		Region:   l.zone,
+		Type:     "g6-nanode-1",
```
rahulait, line 142: nit: can we have this configurable via a variable? Though plans don't change frequently, just wondering if it does get updated in future to g7, we can just set the variable than changing and building again. Same for image.

deploy/chart/values.yaml
```diff
@@ -44,6 +44,11 @@ tolerations:
     operator: Exists
     effect: NoSchedule
 
+# Options for LoadBalancers backed by shared IPs instead of NodeBalancers
+# sharedIPLoadBalancing:
+#   defaultLoadBalancerClass: <linode-nodebalancer or io.cilium/bgp-control-plane>
+#   bgpNodeLabel: <node label for CiliumBGPPeeringPolicy>
+
```
rahulait, line 51: nit: can we also add some documentation regarding these flags in readme?

cloud/linode/client/client.go
```diff
@@ -11,7 +11,12 @@ import (
 type Client interface {
 	GetInstance(context.Context, int) (*linodego.Instance, error)
 	ListInstances(context.Context, *linodego.ListOptions) ([]linodego.Instance, error)
+	CreateInstance(ctx context.Context, opts linodego.InstanceCreateOptions) (*linodego.Instance, error)
 	GetInstanceIPAddresses(context.Context, int) (*linodego.InstanceIPAddressResponse, error)
+	AddInstanceIPAddress(ctx context.Context, linodeID int, public bool) (*linodego.InstanceIP, error)
+	DeleteInstanceIPAddress(ctx context.Context, linodeID int, ipAddress string) error
+
+	ShareIPAddresses(ctx context.Context, opts linodego.IPAddressesShareOptions) error
```
rahulait, line 19: nit: we can remove parameter names and just keep parameter types to keep it consistent with other lines in file. There are some lines which still have both parameter name and types, but in general I see folks trying to just keep parameter types.

AshleyDumaine: If we want to keep it consistent, I would prefer adding parameter names for clarity

cloud/linode/cilium_loadbalancers.go
AshleyDumaine: This is a temporary workaround while there's no way to request IPs that are not attached to a Linode. I'm hoping that changes by the time there is a g7 plan or ubuntu 22.04 goes EOL 😳

cloud/linode/cilium_loadbalancers.go
```diff
@@ -0,0 +1,355 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	newSharedIP, err := l.client.AddInstanceIPAddress(ctx, ipHolder.ID, true)
+	if err != nil {
+		return "", err
+	}
+
+	// need to retrieve existing public IPs on the IP holder since ShareIPAddresses
```
AshleyDumaine, line 82: I think instead here so we can have a single "ip-holder" for multiple clusters, I should just gather up the IPs from the existing (if any) CiliumLoadBalancerPools created by the CCM to determine the full list of IPs to share on the selected nodes

cloud/linode/cilium_loadbalancers.go
```diff
@@ -0,0 +1,445 @@
...
+	ipHolderLabelPrefix        = "linode-ccm-ip-holder"
+	ciliumBGPPeeringPolicyName = "linode-ccm-bgp-peering"
+)
+
+// This mapping is unfortunately necessary since there is no way to get the
+// numeric ID for a data center from the API.
+// These values come from https://www.linode.com/docs/products/compute/compute-instances/guides/failover/#ip-sharing-availability
+var (
+	regionIDMap = map[string]int{
+		"us-southeast": 4,  // Atlanta, GA (USA)
```
tchinmai7, line 38: can we move this to a configmap as a followup?

AshleyDumaine: How would a configmap help? I think we'd still need to cut a new release if we did add/remove a DC, it would just be a difference of if the image version changes vs the k8s resources

tchinmai7: I guess i was thinking more from a internal customer pov - modifying a configmap in place vs picking up a new version - but you're right, doesn't give us much

README.md
```diff
@@ -78,6 +84,43 @@ Key | Values | Default | Description
...
+  name: ccm-linode
+  namespace: kube-system
+spec:
+  template:
+    spec:
+      containers:
+        - image: linode/linode-cloud-controller-manager:latest
+          name: ccm-linode
+          args:
+          - --leader-elect-resource-lock=endpoints
```
rahulait, line 109: nit: example args needs an update as they have changed recently.

AshleyDumaine: updated to just include the args that need to be added for the feature

rahulait (approved):
LGTM, thanks for adding these.

tchinmai7 (approved):

cloud/linode/client/client.go
avestuk: `Make it consistent! By converging on my standard...` 

I love it :D

avestuk (approved):

cloud/linode/cilium_loadbalancers.go
```diff
@@ -0,0 +1,445 @@
...
+			NodeSelector: &slimv1.LabelSelector{MatchLabels: map[string]string{kv[0]: kv[1]}},
+			VirtualRouters: []v2alpha1.CiliumBGPVirtualRouter{{
+				LocalASN:      65001,
+				ExportPodCIDR: ptr.To(true),
+				ServiceSelector: &slimv1.LabelSelector{
+					MatchExpressions: []slimv1.LabelSelectorRequirement{{
+						Key:      "somekey",
+						Operator: slimv1.LabelSelectorOpNotIn,
+						Values:   []string{"never-used-value"},
+					}},
```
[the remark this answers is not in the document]
AshleyDumaine: Yes, this comes from [the cilium documentation](https://docs.cilium.io/en/stable/network/bgp-control-plane/#service-announcements) since by default, virtual routers will not announce services. 
This is a selector that matches everything since it's using "NotIn" with a dummy label.

cloud/linode/cilium_loadbalancers.go
```diff
@@ -0,0 +1,445 @@
...
+				},
+			}},
+		},
+	}
+	for i := 1; i <= 4; i++ {
```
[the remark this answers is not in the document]
AshleyDumaine: Yes, this does, this is copying lelastic: https://github.com/linode/lelastic/blob/master/main.go#L98

cloud/linode/loadbalancers.go
```diff
@@ -184,9 +195,51 @@ func (l *loadbalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+			klog.Infof("Cilium LB IP pool %s for Service %s ensured", pool.Name, serviceNn)
+			// ingress will be set by Cilium
+			return &v1.LoadBalancerStatus{
+				Ingress: service.Status.LoadBalancer.Ingress,
+			}, nil
+		}
+		if err != nil && !k8serrors.IsNotFound(err) {
+			klog.Infof("Failed to get CiliumLoadBalancerIPPool: %s", err.Error())
+			return nil, err
+		}
```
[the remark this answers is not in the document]
AshleyDumaine: Yeah, the comment explaining why is on line 211, we can't simply check for `nil` for the `CiliumLoadBalancerIPPool` for the pool not being found, instead the cilium client will just return an empty `CiliumLoadBalancerIPPool` struct.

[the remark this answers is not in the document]
avestuk: I'm not sure that I follow why that necessitates checking that pool is not nil before the error. 

If the pool does not exist it's not nil, but empty. Does a `IsNotFound` error get returned in this case too?

cloud/linode/cilium_loadbalancers.go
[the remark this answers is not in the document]
AshleyDumaine: added

cloud/linode/cilium_loadbalancers.go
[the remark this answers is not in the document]
AshleyDumaine: added

cloud/linode/loadbalancers.go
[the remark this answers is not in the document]
AshleyDumaine: Ah gotcha, yeah it should be fine to move the error check up earlier. Updated.

srust (approved):

PR #219, opened 2024-06-18
add timeout for linodego http client

rahulait:
### Why this PR is needed
We see cloud-controller-manager stuck for 10-15 mins when doing vpcless installs with firewalling enabled using cluster-api-provider-linode. Upon further troubleshooting, we see some calls getting stuck for more than 10 mins:
```
2024/06/17 13:23:11.262239 DEBUG RESTY
==============================================================================
~~~ REQUEST ~~~
GET  /v4/linode/instances  HTTP/1.1
HOST   : api.linode.com
HEADERS:
	Accept: application/json
	Authorization: Bearer XXXX
	Content-Type: application/json
	User-Agent: linode-cloud-controller-manager linodego/v1.33.0 https://github.com/linode/linodego
BODY   :
***** NO CONTENT *****
------------------------------------------------------------------------------
~~~ RESPONSE ~~~
STATUS       : 200 OK
PROTO        : HTTP/1.1
RECEIVED AT  : 2024-06-17T13:23:11.262101656Z
TIME DURATION: 15m47.632783808s
HEADERS      :
	Access-Control-Allow-Credentials: true
```
This PR adds timeout for http requests so that they complete within the specified time (default 2 min, configurable using LINODE_REQUEST_TIMEOUT environment variable).

#### Steps to reproduce the issue using CAPL:
1. Provision new CAPL cluster
```
CONTROL_PLANE_MACHINE_COUNT=3 WORKER_MACHINE_COUNT=3 FW_AUDIT_ONLY=false clusterctl generate cluster vpcless --infrastructure local-linode:v0.0.0 --flavor kubeadm-vpcless | k apply -f -
```
2. Check time taken by nodes to come up and join the cluster. Third control plane node would have taken more than 12-15mins after the second control plane node joined the cluster.

#### How to test the fix
Provision CAPL cluster for vpcless flavor with linode-ccm image set to image built from this patch. I have a custom image with this patch at `docker.io/rahulait/ccm:timeout`. Once provisioned, check and see that nodes have joined the cluster without much gaps between their provisioning.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/cloud/linode/client/client.go
+++ b/cloud/linode/client/client.go
@@ -4,7 +4,9 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/linode/linodego"
@@ -56,8 +58,11 @@ type Client interface {
 // linodego.Client implements Client
 var _ Client = (*linodego.Client)(nil)
 
-// New creates a new linode client with a given token, userAgent, and API URL
-func New(token, userAgent, apiURL string, timeout time.Duration) (*linodego.Client, error) {
+// New creates a new linode client with a given token and default timeout
+func New(token string, timeout time.Duration) (*linodego.Client, error) {
+	userAgent := fmt.Sprintf("linode-cloud-controller-manager %s", linodego.DefaultUserAgent)
+	apiURL := os.Getenv("LINODE_URL")
+
 	tokenSource := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
 	oauth2Client := &http.Client{
 		Transport: &oauth2.Transport{
@@ -72,6 +77,6 @@ func New(token, userAgent, apiURL string, timeout time.Duration) (*linodego.Clie
 	}
 	client.SetUserAgent(userAgent)
 
-	klog.V(3).Infof("Linode client created with default timeout of %v seconds", timeout)
+	klog.V(3).Infof("Linode client created with default timeout of %v", timeout)
 	return client, nil
 }
--- a/cloud/linode/client/client.go
+++ b/cloud/linode/client/client.go
@@ -4,8 +4,17 @@ package client
 
 import (
 	"context"
+	"net/http"
+	"time"
 
 	"github.com/linode/linodego"
+	"golang.org/x/oauth2"
+	"k8s.io/klog/v2"
+)
+
+const (
+	// DefaultClientTimeout is the default timeout for a client Linode API call
+	DefaultClientTimeout = 120 * time.Second
 )
 
 type Client interface {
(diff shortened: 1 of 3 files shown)


cloud/linode/cloud.go
```diff
@@ -98,7 +100,15 @@ func newCloud() (cloudprovider.Interface, error) {
 	url := os.Getenv(urlEnv)
 	ua := fmt.Sprintf("linode-cloud-controller-manager %s", linodego.DefaultUserAgent)
 
-	linodeClient, err := client.New(apiToken, ua, url)
+	// set timeout used by linodeclient for API calls
+	timeout := client.DefaultClientTimeout
+	if raw, ok := os.LookupEnv("LINODE_REQUEST_TIMEOUT"); ok {
```
AshleyDumaine, line 105: We want to rename this var to indicate the time unit?

rahulait: Updated.

cloud/linode/client/client.go
```diff
@@ -48,14 +57,22 @@ type Client interface {
...
+		},
+		Timeout: timeout,
+	}
+	linodeClient := linodego.NewClient(oauth2Client)
 	client, err := linodeClient.UseURL(apiURL)
 	if err != nil {
 		return nil, err
 	}
 	client.SetUserAgent(userAgent)
 	client.SetToken(token)
```
AshleyDumaine, line 74: I think we can remove this line if we're setting the token in the oauth2 client?

rahulait: Updated

AshleyDumaine (approved):
LGTM

cloud/linode/client/client.go
```diff
@@ -48,14 +57,21 @@ type Client interface {
 var _ Client = (*linodego.Client)(nil)
 
 // New creates a new linode client with a given token, userAgent, and API URL
-func New(token, userAgent, apiURL string) (*linodego.Client, error) {
-	linodeClient := linodego.NewClient(nil)
+func New(token, userAgent, apiURL string, timeout time.Duration) (*linodego.Client, error) {
```
nesv, line 60: :question: Is it necessary to pass in `userAgent` and `apiURL`, or could those be safely defaulted?

If the use-case for specifying these is to override them in testing scenarios, that should be a separate concern that is handled in the testing code.

cloud/linode/cloud.go
```diff
@@ -98,7 +100,15 @@ func newCloud() (cloudprovider.Interface, error) {
 	url := os.Getenv(urlEnv)
 	ua := fmt.Sprintf("linode-cloud-controller-manager %s", linodego.DefaultUserAgent)
 
-	linodeClient, err := client.New(apiToken, ua, url)
+	// set timeout used by linodeclient for API calls
+	timeout := client.DefaultClientTimeout
+	if raw, ok := os.LookupEnv("LINODE_REQUEST_TIMEOUT_SECONDS"); ok {
+		if t, _ := strconv.Atoi(raw); t > 0 {
+			timeout = time.Duration(t) * time.Second
+		}
```
nesv, line 108: nit: This reads a little strangely to me.

Instead, I would recommend keying off of the returned error to know whether or not `t` is usable:

```suggestion
		if t, err := strconv.Atoi(raw); err == nil {
			timeout = time.Duration(t) * time.Second
		}
```

NOTE: Could the `LINODE_REQUEST_TIMEOUT_SECONDS` environment variable be renamed to `LINODE_REQUEST_TIMEOUT` and hold a duration string that is parsed with [time.ParseDuration](https://pkg.go.dev/time#ParseDuration)?

cloud/linode/client/client.go
rahulait: No, they can be safely defaulted. Updated the PR to use default values.

cloud/linode/cloud.go
rahulait: Updated the code. We could rename the `LINODE_REQUEST_TIMEOUT_SECONDS` as suggested, just trying to keep the code same as we have at other places, hence keeping it as it is here. Let me know if you still feel we should use time.ParseDuration and I can update this codepath as well.

nesv: > just trying to keep the code same as we have at other places

That's fair! It was just an idea. :smile:

> Let me know if you still feel we should use time.ParseDuration and I can update this codepath as well.

No, I don't think this is necessary.

nesv (approved):

cloud/linode/client/client.go
nesv: Sweet! Thank you.

PR #225, opened 2024-08-14
fix: refresh node metadata on stale nodes

shanduur:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

Sometimes the metadata might become stale when the node itself is not removed/added or even updated. We need to trigger the update to prevent stale metadata on the nodes.

### DIFF
--- a/cloud/linode/node_controller.go
+++ b/cloud/linode/node_controller.go
@@ -23,6 +23,11 @@ import (
 	"github.com/linode/linode-cloud-controller-manager/cloud/linode/client"
 )
 
+const (
+	informerResyncPeriod = 1 * time.Minute
+	defaultMetadataTTL   = 300 * time.Second
+)
+
 type nodeController struct {
 	sync.RWMutex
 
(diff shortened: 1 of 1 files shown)


cloud/linode/node_controller.go
```diff
@@ -57,17 +57,29 @@ func newNodeController(kubeclient kubernetes.Interface, client client.Client, in
...
+				if !ok {
+					return
+				}
+
+				klog.Infof("NodeController will handle newly updated node (%s) metadata", node.Name)
+				s.queue.Add(node)
+			},
 		},
-	}); err != nil {
+		5*time.Minute,
```
okokes-akamai, line 81: It's a bit lost here, can we maybe make that a constant up top, so that we know about these global refresh periods and have them in one place?

okokes-akamai: Also - the five minutes exactly the same as the TTL for the node refresh, so it's a bit racey (depending on how long the actions take).

shanduur: Fixed, and reduced the informer timeout to 1 minute.

okokes-akamai: Cool. It's gonna be a bit spammy - it'll say "not updating due to TTL" four times in a row and then updating once, but I guess it's hard to align these time frames, especially as they'll drift anyway (due to the routines taking some time and also given that these intervals are not guaranteed to be kept exactly).

okokes-akamai (approved):

rammanoj (approved):

PR #226, opened 2024-08-15
Bump golangci-lint to support 1.23

okokes-akamai:
We use 'stable' in our Github Actions and that now refers to Go 1.23, so we need a supported version of golangci-lint for that one to work.

### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,9 +25,9 @@ vet: fmt
 .PHONY: lint
 lint:
 	docker run --rm -v "$(shell pwd):/var/work:ro" -w /var/work \
-		golangci/golangci-lint:v1.57.2 golangci-lint run -v --timeout=5m
+		golangci/golangci-lint:v1.60.1 golangci-lint run -v --timeout=5m
 	docker run --rm -v "$(shell pwd):/var/work:ro" -w /var/work/e2e \
-		golangci/golangci-lint:v1.57.2 golangci-lint run -v --timeout=5m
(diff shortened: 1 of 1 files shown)


okokes-akamai:
Oh... it seems to have worked, it was just slow (as it was the first run). https://github.com/linode/linode-cloud-controller-manager/actions/runs/10399222522/job/28797786593

okokes-akamai:
I'd rather move golangci-lint to tools.go and go.mod, so that dependabot handles this automatically.

rahulait:
Since this has already been updated in main branch and we are now using `golangci/golangci-lint:latest`, can we close this PR?

okokes-akamai:
@rahulait we're using `latest`? That sounds like a bad idea that will only introduce problems after an upstream release breaks our build.

rahulait:
If upstream breaks, we can easily pin to an older release for getting past linting errors. Do you think its better to pin to a fixed version or use go.mod and have dependabot update it whenever required?

okokes-akamai:
The issue is that if it breaks, it just breaks over time and all existing PRs and nightly builds and whatnot will all be red. I don't think we need to go all the way to `go.mod` pinning, but we could just pin it in the job itself and use dependabot to track it.

rahulait:
I am hoping this will address the issue: https://github.com/linode/linode-cloud-controller-manager/pull/278
Let me know if this is not what we wanted.

okokes-akamai:
Great!

PR #233, opened 2024-09-23
Get IP addresses type from Linode

Michkov:
Avoid auto-detection of private IP address - respect setting in Linode.
Needed to support testing Linode instances which are providing public IP addresses in private IP ranges. 

### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 

### DIFF
--- a/cloud/linode/instances.go
+++ b/cloud/linode/instances.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"slices"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


rahulait:
We had similar implementation in the past where we were making linode API call to get instance ip addresses for each instance (`client.GetInstanceIPAddresses(ctx, instance.ID)`) but that ended up in an escalation. If we try to get ip-address for each instance, then that would mean accounts which have large number of instances will keep on making those many additional calls to linode API every 15-20 secs when the cache expires. Hence we rely on one call (ListInstances) and build cache from the response.

rahulait (changes requested):
We can't make additional calls for each instance everytime we build the cache as it balloons up quickly.

Michkov:
Replaced by #235

Issue #111, opened 2022-11-30
Error When Deploying CCM On New Cluster


lots0logs:
## General:

* [x] Have you removed all sensitive information, including but not limited to access keys and passwords?
* [x] Have you checked to ensure there aren't other open or closed [Pull Requests](../../pulls) for the same bug/feature/question?

----

## Bug Reporting
I cannot get CCM to work. The pod runs but there is an error in the logs that keeps repeating itself indefinitely.

### Expected Behavior
CCM should work without error.

### Steps to Reproduce the Problem

  1. Deploy CCM onto a new single node RKE cluster
  2. Check logs of the CCM pod

### Environment Specifications
Kubernetes: 1.21.7
RKE: 1.3.15

#### Screenshots, Code Blocks, and Logs
```
E1130 18:33:49.503962       1 node_controller.go:212] error syncing 'rancher-dev': failed to get node modifiers from cloud provider: failed to find kubelet node IP from cloud provider, requeuing
2022/11/30 18:34:22.375645 DEBUG RESTY
==============================================================================
~~~ REQUEST ~~~
GET  /v4/linode/instances/redacted  HTTP/1.1
HOST   : api.linode.com
HEADERS:
Accept: application/json
Authorization: Bearer redacted
Content-Type: application/json
User-Agent: linode-cloud-controller-manager linodego/v0.32.2 https://github.com/linode/linodego
BODY   :
***** NO CONTENT *****
------------------------------------------------------------------------------
~~~ RESPONSE ~~~
STATUS       : 200 OK
RECEIVED AT  : 2022-11-30T18:34:22.375520522Z
TIME DURATION: 95.928433ms
HEADERS      :
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: Authorization, Origin, X-Requested-With, Content-Type, Accept, X-Filter
Access-Control-Allow-Methods: HEAD, GET, OPTIONS, POST, PUT, DELETE
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Status
Cache-Control: private, max-age=0, s-maxage=0, no-cache, no-store, private, max-age=60, s-maxage=60
Connection: keep-alive
Content-Length: 736
Content-Security-Policy: default-src 'none'
Content-Type: application/json
Date: Wed, 30 Nov 2022 18:34:22 GMT
Retry-After: 27
Server: nginx
[...]

PR #236, opened 2024-10-01
[fix] handle cases where the internal-ip is already set (e.g. kubelet)

AshleyDumaine:
Related to #111 

### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 

In the case that a Node's internal IP is already set (e.g. to a VLAN IP), the CCM will fail to set the external IP for the Node:
```
E1001 18:20:24.780329       1 node_controller.go:240] error syncing 'foo-control-plane-kxlgm': failed to get node modifiers from cloud provider: provided node ip for node "foo-control-plane-kxlgm" is not valid: failed to get node address from cloud provider that matches ip: 10.0.0.1, requeuing
```

### DIFF
--- a/cloud/linode/instances.go
+++ b/cloud/linode/instances.go
@@ -285,6 +285,13 @@ func (i *instances) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloud
 		addresses = append(addresses, v1.NodeAddress{Type: ip.ipType, Address: ip.ip})
 	}
 
+	// include IPs set by kubelet for internal node IP
+	for _, addr := range node.Status.Addresses {
+		if addr.Type == v1.NodeInternalIP {
+			addresses = append(addresses, v1.NodeAddress{Type: v1.NodeInternalIP, Address: addr.Address})
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


rosskirkpat (approved):
LGTM

cloud/linode/instances.go
```diff
@@ -285,6 +285,13 @@ func (i *instances) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloud
 		addresses = append(addresses, v1.NodeAddress{Type: ip.ipType, Address: ip.ip})
 	}
 
+	// include IPs set by kubelet for internal node IP
+	for _, addr := range node.Status.Addresses {
+		if addr.Type == v1.NodeInternalIP {
+			addresses = append(addresses, v1.NodeAddress{Type: v1.NodeInternalIP, Address: addr.Address})
+		}
+	}
```
rahulait, line 293: Should we make sure same ip is not already present in the slice? For example, if this InstanceMetadata() gets called again after 5-10 mins, then we'll end up adding all the existing ip's again into the slice. Not sure if k8s automatically handles such case.

rahulait (approved):
lgtm

PR #276, opened 2024-12-17
[fix] : use single copy of instances cache

rahulait:
This PR allows CCM to use single copy of instance cache for all controllers. Previously, all controllers were maintaining their own copy of instance cache and were refreshing it when using it.

To test, I intentionally started CCM with flags `- --node-status-update-frequency=20s` and `- --linodego-debug=true` which makes CCM to update node status every 20s rather than [default 5min](https://github.com/kubernetes/cloud-provider/blob/release-1.32/config/v1alpha1/defaults.go#L37). With existing main branch and mentioned flags, I can see it making double calls to linode API for updating cache (one for instance cache kept by route_controller and one by instances controller). With an image from this PR, I can see it makes half of what I see with main branch. It confirms that both controllers are now sharing the same cache.
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 


cloud/linode/cloud.go
```diff
@@ -123,7 +125,7 @@ func newCloud() (cloudprovider.Interface, error) {
 	// create struct that satisfies cloudprovider.Interface
 	lcloud := &linodeCloud{
```
okokes-akamai, line 126: The `Initialize` method here calls `newInstances` yet again for the Node Controller - do we want that one to share this `instances` struct, too?

rahulait: Thank you, good catch. Updated the PR to address this as well.

AshleyDumaine (approved):

PR #293, opened 2025-01-13
[feat] Linode Client metrics

rsienko:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 
1. [x] Have you linked your PR to an [issue](https://github.com/linode/linode-cloud-controller-manager/pull/296)


rsienko:
Example new metrics:
```
# HELP client_operations_total client counters for each operation and its result
# TYPE client_operations_total counter
client_operations_total{method="ListInstances",result="ok"} 1
client_operations_total{method="ListNodeBalancerConfigs",result="ok"} 2
client_operations_total{method="ListNodeBalancerFirewalls",result="ok"} 2
client_operations_total{method="ListNodeBalancerNodes",result="ok"} 2
client_operations_total{method="ListNodeBalancers",result="ok"} 2
client_operations_total{method="ListVPCIPAddresses",result="ok"} 1
client_operations_total{method="ListVPCs",result="ok"} 1
client_operations_total{method="RebuildNodeBalancerConfig",result="ok"} 2
```

rsienko:
I didn't find a good way to fix coverage for the generated file.

It doesn't look trivial to generate calls to all the methods. Gowrap provides method declaration as a string. I'd need to parse the Interface from AST. Doable but is it worth it?

A generated test case would look like so and it's pretty much all boilerplate.

```
func TestAddInstanceIPAddress(t *testing.T) {
	ctrl := gomock.NewController(t)
	defer ctrl.Finish()

	mock := mocks.NewMockClient(ctrl)
	expIP := linodego.InstanceIP{}
	var m1 = &dto.Metric{}
	var m2 = &dto.Metric{}

	client := NewClientWithPrometheus(mock)
	mock.EXPECT().AddInstanceIPAddress(gomock.Any(), 1, true).DoAndReturn(
		func(ctx context.Context, id int, isActive bool) (*linodego.InstanceIP, error) {
			return &expIP, nil
		})
	ec, err := ClientMethodCounterVec.GetMetricWithLabelValues("AddInstanceIPAddress", resultOk)
	if err != nil {
		t.Error(err)
	}

	ec.Write(m1)
	retIP, err := client.AddInstanceIPAddress(context.TODO(), 1, true)
	ec.Write(m2)

	assert.Equal(t, m1.Counter.GetValue()+1, m2.Counter.GetValue(), "'ok' counter should be incremented")
	assert.Equal(t, expIP, *retIP, "IP should be returned unmodified")
	assert.Nil(t, err, "nil error should be returned unmodified")
}
```

Note that `Client` interface methods expose `error` returned in case call fails with no response. Thus, an approach using Resty's callbacks would be incomplete.

```
client.OnAfterResponse(clientMetrics)
...
func clientMetrics(response *resty.Response) (err error) {
	code := response.StatusCode()
	ClientMethodCounterVec.WithLabelValues("CreateFirewallDevice", strconv.Itoa(code)).Inc()
}
```

Note, it also brings other aspects - retries and path patterns into label cardinality consideration (we cannot expose resource IDs). Example form linodego:

```
func (c *Client) ListNodeBalancerFirewalls(ctx context.Context, nodebalancerID int, opts *ListOptions) ([]Firewall, error) {
  response, err := getPaginatedResults[Firewall](ctx, c, formatAPIPath("nodebalancers/%d/firewalls", nodebalancerID), opts)
  ...
```

PR #295, opened 2025-01-15
[feat] add linode token health check

ficap-akamai:
Sometimes it happens that the LINODE API TOKEN used by ccm expires or is revoked. This can go unnoticed for long periods since the only mechanism in place is logging it when actual call to Linode API fails. This PR implements health checker component that periodically makes calls to authenticated linode api endpoint (/v4/profile) to determine validity of token being used. Upon detection of invalid token health checker sends a signal to shut down controllers. This results in ccm Pod CrashLoopBackOffing which is greatly visible and can be acted upon. Health checker is only started when the controller is elected the leader. This reduces number of calls done to Linode API if there are multiple replicas of ccm. Whole health checker functionality is behind `--enable-token-health-checker` flag to be explicitly enabled if needed rather than enabling it by default.

### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 

### DIFF
--- a/cloud/linode/client/client.go
+++ b/cloud/linode/client/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/linode/linodego"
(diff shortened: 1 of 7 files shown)


rahulait:
@ficap-akamai if you have write access to the repo, can you open the PR directly from a branch in original repo? It seems like GHAs are failing if PR is coming from a forked repo.

ficap-akamai:
closing in favor of https://github.com/linode/linode-cloud-controller-manager/pull/296

PR #296, opened 2025-01-16
[feat] add linode token health check

ficap-akamai:
Sometimes it happens that the LINODE API TOKEN used by ccm expires or is revoked. This can go unnoticed for long periods since the only mechanism in place is logging it when actual call to Linode API fails. This PR implements health checker component that periodically makes calls to authenticated linode api endpoint (/v4/profile) to determine validity of token being used. Upon detection of invalid token health checker sends a signal to shut down controllers. This results in ccm Pod CrashLoopBackOffing which is greatly visible and can be acted upon. Health checker is only started when the controller is elected the leader. This reduces number of calls done to Linode API if there are multiple replicas of ccm. Whole health checker functionality is behind `--enable-token-health-checker` flag to be explicitly enabled if needed rather than enabling it by default.

### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 

### DIFF
--- a/cloud/linode/client/client.go
+++ b/cloud/linode/client/client.go
@@ -4,6 +4,7 @@ package client
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -52,6 +53,8 @@ type Client interface {
 	DeleteFirewall(ctx context.Context, fwid int) error
 	GetFirewall(context.Context, int) (*linodego.Firewall, error)
 	UpdateFirewallRules(context.Context, int, linodego.FirewallRuleSet) (*linodego.FirewallRuleSet, error)
+
+	GetProfile(ctx context.Context) (*linodego.Profile, error)
 }
 
 // linodego.Client implements Client
@@ -72,4 +75,22 @@ func New(token string, timeout time.Duration) (*linodego.Client, error) {
 
 	klog.V(3).Infof("Linode client created with default timeout of %v", timeout)
 	return client, nil
+}
+
+func CheckClientAuthenticated(ctx context.Context, client Client) (bool, error) {
+	_, err := client.GetProfile(ctx)
+	if err == nil {
+		return true, nil
+	}
+
+	var linodeErr *linodego.Error
+	if !errors.As(err, &linodeErr) {
+		return false, err
+	}
+
+	if linodego.ErrHasStatus(err, http.StatusUnauthorized) {
+		return false, nil
+	}
+
+	return false, err
 }
--- a/cloud/linode/cloud.go
+++ b/cloud/linode/cloud.go
@@ -1,6 +1,7 @@
 package linode
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"net"
@@ -19,11 +20,12 @@ import (
 
 const (
 	// The name of this cloudprovider
-	ProviderName       = "linode"
-	accessTokenEnv     = "LINODE_API_TOKEN"
-	regionEnv          = "LINODE_REGION"
-	ciliumLBType       = "cilium-bgp"
-	nodeBalancerLBType = "nodebalancer"
+	ProviderName           = "linode"
+	accessTokenEnv         = "LINODE_API_TOKEN"
+	regionEnv              = "LINODE_REGION"
+	ciliumLBType           = "cilium-bgp"
+	nodeBalancerLBType     = "nodebalancer"
+	tokenHealthCheckPeriod = 5 * time.Minute
 )
 
 var supportedLoadBalancerTypes = []string{ciliumLBType, nodeBalancerLBType}
@@ -32,9 +34,10 @@ var supportedLoadBalancerTypes = []string{ciliumLBType, nodeBalancerLBType}
 // We expect it to be initialized with flags external to this package, likely in
 // main.go
 var Options struct {
-	KubeconfigFlag        *pflag.Flag
-	LinodeGoDebug         bool
-	EnableRouteController bool
+	KubeconfigFlag           *pflag.Flag
+	LinodeGoDebug            bool
+	EnableRouteController    bool
+	EnableTokenHealthChecker bool
 	// Deprecated: use VPCNames instead
 	VPCName               string
 	VPCNames              string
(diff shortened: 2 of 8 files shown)


cloud/linode/health_check_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+func testSucceedingCallsToLinodeAPIHappenStopSignalNotFired(t *testing.T, client *linodego.Client, api *fakeAPI) {
+	writableStopCh := make(chan struct{})
+	readableStopCh := make(chan struct{})
+	const validToken = "validtoken"
+
+	api.tkn = validToken
+	client.SetToken(validToken)
+
+	hc, err := newHealthChecker(validToken, 1*time.Second, 1*time.Second, writableStopCh)
```
rsienko, line 49: Extra API call every second creates a risk that Linode account gets blocked by overloading the API. How about doing it once at startup and once a day to detect expired tokens?

rsienko: See also this [PR draft](https://github.com/linode/linode-cloud-controller-manager/pull/293) which introduces metrics that can expose client errors at realtime, including token related.

cloud/linode/cloud.go
```diff
@@ -91,6 +95,24 @@ func newCloud() (cloudprovider.Interface, error) {
...
+		authenticated, err := client.CheckClientAuthenticated(context.TODO(), linodeClient)
+		if err != nil {
+			return nil, fmt.Errorf("linode client authenticated connection error: %w", err)
+		}
+
+		if !authenticated {
+			return nil, fmt.Errorf("linode api token '%s' is invalid", accessTokenEnv)
+		}
+
+		healthChecker, err = newHealthChecker(apiToken, timeout, time.Minute, Options.GlobalStopChannel)
```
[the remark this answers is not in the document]
ficap-akamai: I think that doing the check once a day is rather impractical in this case since invalid token may become unnoticed for the whole day in the worst case. I also think that we should balance length of the period to be long enough to not overload Linode API and at the same time be short enough to be practical.
When I started working on this feature I didn't know that adding metrics was planned. I think that in the future these two can be combined to accomplish desired result (notify user of invalid token) in both ways - user will be able to build their own alerting rules based on new metrics or opt for built in health check (that can use intercepted metrics data to enhance its behavior lowering count of Linode API calls made).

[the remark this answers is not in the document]
akaokunc: Not sure if it is a good idea to do the check every minute. It seems to be too often from my point of view. If you imagine number of deployed CCMs it would be quite significant load caused by this. What if we do it with some longer period (>=1hr) to make sure it is being noticed even if we are not doing any API calls and together with that we will be checking returned values from API calls and we could crash on getting the first 401 from any API call.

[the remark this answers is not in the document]
okokes-akamai: To add a bit of context - we list the customer's Linodes every five minutes - an order of magnitude more expensive operation (that is not even size constrained given what it does). So I'd be more than happy to get the customer profile every minute or five.

cloud/linode/client/client.go
```diff
@@ -73,3 +76,16 @@ func New(token string, timeout time.Duration) (*linodego.Client, error) {
...
 	return client, nil
 }
+
+func CheckClientAuthenticated(ctx context.Context, client Client) (bool, error) {
+	_, err := client.GetProfile(ctx)
+	if err == nil {
+		return true, nil
+	}
+
+	if strings.Contains(err.Error(), "Invalid Token") {
```
eljohnson92, line 86: are there any linodeGo error types we could use here instead of using string parsing for the error message?

cloud/linode/cloud.go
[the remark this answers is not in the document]
rahulait: +1 for doing it every 5 mins

cloud/linode/fake_linode_test.go
```diff
@@ -674,6 +676,29 @@ func (f *fakeAPI) setupRoutes() {
 		rr, _ := json.Marshal(resp)
 		_, _ = w.Write(rr)
 	})
+
+	f.mux.HandleFunc("GET /v4/profile", func(w http.ResponseWriter, r *http.Request) {
```
rahulait, line 680: Instead of adding it to fakeAPI, can we use this in tests? In that case, we don't need this change.
```
ctrl := gomock.NewController(t)
defer ctrl.Finish()
client := mocks.NewMockClient(ctrl)
client.EXPECT().GetProfile(gomock.Any()).Times(1).Return(&linodego.Profile{}, nil)
```

cloud/linode/cloud.go
```diff
@@ -32,9 +33,10 @@ var supportedLoadBalancerTypes = []string{ciliumLBType, nodeBalancerLBType}
 // We expect it to be initialized with flags external to this package, likely in
 // main.go
 var Options struct {
-	KubeconfigFlag        *pflag.Flag
-	LinodeGoDebug         bool
-	EnableRouteController bool
+	KubeconfigFlag           *pflag.Flag
+	LinodeGoDebug            bool
+	EnableRouteController    bool
+	EnableTokenHealthChecker bool
```
rahulait, line 39: This flag should also be added to helm chart:
https://github.com/linode/linode-cloud-controller-manager/blob/main/deploy/chart/templates/daemonset.yaml
https://github.com/linode/linode-cloud-controller-manager/blob/main/deploy/chart/values.yaml

cloud/linode/client/client.go
ficap-akamai: I found something that could work. Please take a look.

cloud/linode/client/client.go
```diff
@@ -73,3 +76,21 @@ func New(token string, timeout time.Duration) (*linodego.Client, error) {
...
+func CheckClientAuthenticated(ctx context.Context, client Client) (bool, error) {
+	_, err := client.GetProfile(ctx)
+	if err == nil {
+		return true, nil
+	}
+
+	var linodeErr *linodego.Error
+	if !errors.As(err, &linodeErr) {
+		return false, err
+	}
```
rahulait, line 89:
```suggestion

```
Wondering if this would make sense as we are anyhow returning `false, err` in the end.

cloud/linode/client/client.go
```diff
@@ -4,6 +4,7 @@ package client
 
 import (
 	"context"
+	"errors"
```
rahulait, line 7:
```suggestion

```
To support below comment

rahulait:
Other than minor nit comment, the PR looks good to me. If you can rebase it with latest main, I'll revisit and add the approvals.

cloud/linode/client/client.go
ficap-akamai: I put it there since `linodego.ErrHasStatus(err, http.StatusUnauthorized)` returns also false if passed error is not `linodego.Error` (https://github.com/linode/linodego/blob/7d14d5c417f97befe1a43d777297bf79d6c15d1c/errors.go#L213). I wanted to distinguish these two situations because when we cannot reliably detect Unauthorized from the err we should return `false, err` instead of `false, nil` like it would happen if this check was not there.

rahulait (approved):
LGTM

PR #307, opened 2025-01-28
[fix] : delete fw if provisioned by CCM with ACL

rahulait:
### General:
Previously, on deletion of service which had ACL annotation for firewall config, those firewalls created by CCM were not cleaned up. This PR adds logic to clean such firewalls created by CCM.

### Pull Request Guidelines:

1. [x] Have you added tests? 

### DIFF
--- a/cloud/linode/loadbalancers.go
+++ b/cloud/linode/loadbalancers.go
@@ -558,6 +558,12 @@ func (l *loadbalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterNa
 		return nil
 	}
 
+	fwClient := firewall.LinodeClient{Client: l.client}
+	err = fwClient.DeleteNodeBalancerFirewall(ctx, service, nb)
+	if err != nil {
+		return err
+	}
+
 	if err = l.client.DeleteNodeBalancer(ctx, nb.ID); err != nil {
 		klog.Errorf("failed to delete NodeBalancer (%d) for service (%s): %s", nb.ID, serviceNn, err)
 		sentry.CaptureError(ctx, err)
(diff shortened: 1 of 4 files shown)


cloud/linode/loadbalancers.go
```diff
@@ -558,6 +558,12 @@ func (l *loadbalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterNa
 		return nil
 	}
 
+	fwClient := firewall.LinodeClient{Client: l.client}
+	err = fwClient.DeleteNodeBalancerFirewall(ctx, service, nb)
```
tchinmai7, line 562: nit
```suggestion
	if err = fwClient.DeleteNodeBalancerFirewall(ctx, service, nb); err != nil {
```

rahulait: updated.

cloud/linode/loadbalancers.go
```diff
@@ -558,6 +558,11 @@ func (l *loadbalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterNa
 		return nil
 	}
 
+	fwClient := firewall.LinodeClient{Client: l.client}
+	if err = fwClient.DeleteNodeBalancerFirewall(ctx, service, nb); err != nil {
```
tchinmai7, line 562: sorry i missed this earlier 😅 
```suggestion
	if err := fwClient.DeleteNodeBalancerFirewall(ctx, service, nb); err != nil {
```

rahulait: err var was defined previously at line 536 https://github.com/linode/linode-cloud-controller-manager/blob/remove-fw/cloud/linode/loadbalancers.go#L536 , so should be fine with = instead of := . Do we really need : here?

tchinmai7: not a big deal - but its good to scope the err's within the if, especially if we're returning right after checking that - this is to avoid shadowing stuff if we have defer's (Which we don't)

tchinmai7 (approved):

cloud/linode/loadbalancers.go
rahulait: I see, yeah that makes sense.

PR #328, opened 2025-02-18
[improvement] Add subnet filtering for routes management

Andrews2024:
### General:
Currently, route_controller manages and configures routes for all Linodes within a VPC if specified. This change adds the option to specify that CCM manages only nodes within a particular subnet or list of subnets.

### Pull Request Guidelines:

1. [x] Have you added tests? 


main.go
```diff
@@ -84,6 +84,7 @@ func main() {
 	command.Flags().BoolVar(&linode.Options.EnableTokenHealthChecker, "enable-token-health-checker", false, "enables Linode API token health checker")
 	command.Flags().StringVar(&linode.Options.VPCName, "vpc-name", "", "[deprecated: use vpc-names instead] vpc name whose routes will be managed by route-controller")
 	command.Flags().StringVar(&linode.Options.VPCNames, "vpc-names", "", "comma separated vpc names whose routes will be managed by route-controller")
+	command.Flags().StringVar(&linode.Options.SubnetNames, "subnet-names", "", "comma separated subnet names whose routes will be managed by route-controller (requires vpc-names flag to also be set)")
```
rahulait, line 87: This change needs to be also added to helm charts here:
https://github.com/linode/linode-cloud-controller-manager/blob/main/deploy/chart/templates/daemonset.yaml#L39

Andrews2024: Updated here: https://github.com/linode/linode-cloud-controller-manager/pull/328/files#diff-14d9977d260e8d6ab9fe6366a5d79684839c794d8310964a724a0f8dd67dcbda

cloud/linode/vpc.go
```diff
@@ -59,13 +75,68 @@ func GetVPCID(ctx context.Context, client client.Client, vpcName string) (int, e
...
+		// subnetIDList is a slice of strings for ease of use with resultFilter
+		subnetNames := strings.Split(Options.SubnetNames, ",")
+		subnetIDList := []string{}
+
+		for _, name := range subnetNames {
+			// For caching
+			subnetID, err := GetSubnetID(ctx, client, vpcID, name)
+			// Don't filter subnets we can't find
+			if err != nil {
+				klog.Errorf("subnet %s not found. Skipping.", name)
```
rahulait, line 123: We should also log the error string to know why it was not found.

deploy/chart/templates/daemonset.yaml
```diff
@@ -60,6 +60,9 @@ spec:
             {{- if not (or $vpcName $vpcNames) }}
             {{- fail "Neither vpcName nor vpcNames is set. Please set one of them." }}
             {{- end }}
+            {{- with .Values.routeController.subnetNames}}
+            - --subnet-names={{ . }}
+            {{- end }}
```
rahulait, line 65: We also need to add this to values.yaml file. Also, since we now allow specifying vpcNames outside of routeController block, we should allow specifying subnetNames there as well. Try checking the logic for vpcNames in the daemonset.yaml file and update accordingly.
Additional file which needs update:
https://github.com/linode/linode-cloud-controller-manager/blob/main/deploy/chart/values.yaml#L79

Andrews2024: Updated here (https://github.com/linode/linode-cloud-controller-manager/pull/328/files#diff-46ba93a79d17b3b2e6b5b957292c18110b19e37f46c2a66827688feb77ecd831) and here (https://github.com/linode/linode-cloud-controller-manager/pull/328/files#diff-14d9977d260e8d6ab9fe6366a5d79684839c794d8310964a724a0f8dd67dcbda)

rahulait (approved):
LGTM 🚀

PR #332, opened 2025-02-20
[improvement] Allow setting vpcName(s) without enabling route controller

cbbond:
### General:

### Pull Request Guidelines:

1. [ ] Have you added tests? 

This PR allows Helm chart consumers to set `vpcName(s)` without enabling the route controller. The chart will still error if `routeController` is set to true but no VPC is specified. This change allows the node controller to only cache nodes that belong to given VPCs without enabling route controller functionality.

### DIFF
--- a/deploy/chart/templates/daemonset.yaml
+++ b/deploy/chart/templates/daemonset.yaml
@@ -44,25 +44,33 @@ spec:
             {{- with .Values.linodegoDebug }}
             - --linodego-debug={{ . }}
             {{- end }}
-            {{- if .Values.routeController }}
-            - --enable-route-controller=true
-            {{- if and .Values.routeController.vpcName .Values.routeController.vpcNames }}
+            {{- $vpcNames := .Values.vpcNames }}
+            {{- if and .Values.routeController .Values.routeController.vpcNames }}
+            {{- $vpcNames = .Values.routeController.vpcNames }}
+            {{- end }}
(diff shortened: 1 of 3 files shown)


cbbond:
> The changes look good to me. Just wondering if it should be marked a breaking change or not as others using route-controller will now have to make sure their values.yaml is also updated than just bumping the release version.

@rahulait That makes sense to me, I'll add the breaking label to the PR.

rahulait (approved):

rosskirkpat:
> The changes look good to me. Just wondering if it should be marked a breaking change or not as others using route-controller will now have to make sure their values.yaml is also updated than just bumping the release version.

Good point, we should add in backwards compatibility to the templating logic to avoid having this be a breaking change. Thinking something like this at the top of daemonset.yaml should work

```yaml
{{- $vpcNames := .Values.vpcNames }}
{{- if .Values.routeController.vpcNames }}
{{- $vpcNames := Values.routeController.vpcNames }}
{{- end }}
{{- $vpcName := .Values.vpcName }}
{{- if .Values.routeController.vpcName }}
{{- $vpcNames := Values.routeController.vpcName }}
{{- end }}
{{- if and $vpcNames $vpcName }}
{{- fail "Both vpcName and vpcNames are set. Please use only vpcNames." }}
{{- end }}
---
```

PR #341, opened 2025-03-03
[fix] : use latest node object when processing node update request

rahulait:
### General:
This PR now makes sure most recent node request is processed. Currently, in case of 429 or 5XX errors, we are re-queueing the request after a minute to retry again. If linodeAPI is down for some reason or we are getting rate limited, those requests will get re-queued. We also get node update requests every 5 mins and if they also fail, they also end up getting queued. We saw that we were making multiple requests again and again which kept on increasing over time when we encountered 429 or 5XX API errors. This PR addresses that and keeps track of time when the most recent node update request was made for a node. All prior requests which are in queue or are waiting to be queued due to AddAfter() will get ignored once it tries to process them and will always process the most recent request. 

Here is the memory usage graph for CCM before and after fix (after intentionally making CCM to requeue requests to mimic behavior on 429 or 5XX errors):
Screenshot 2025-03-08 at 2 40 23 PM

### Pull Request Guidelines:

1. [x] Have you added tests? 

### DIFF
--- a/cloud/linode/instances_test.go
+++ b/cloud/linode/instances_test.go
@@ -17,7 +17,6 @@ import (
 	cloudprovider "k8s.io/cloud-provider"
 
 	"github.com/linode/linode-cloud-controller-manager/cloud/linode/client/mocks"
-
 )
 
 const (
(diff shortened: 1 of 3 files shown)


AshleyDumaine (approved):

komer3 (approved):
lgtm!

PR #345, opened 2025-03-06
[CI] Golangci-lint update part 3

Andrews2024:
### General:
This is the third update to the linting configuration for the repository. This resolves errors for the following linters:
* forcetypeassert
* govet
* thelper

After attempting to resolve errors raised by paralleltest, many tests broke completely or were flaky. Therefore, I'm leaving paralleltest out of the configuration with a note on why.

### Pull Request Guidelines:


.github/workflows/build-test.yml
```diff
@@ -74,9 +76,7 @@ jobs:
         run: make vet
 
       - name: lint
-        uses: golangci/golangci-lint-action@v6
-        with:
-          version: v1.64.5
+        run: make lint
```
rahulait, line 79: `make lint` uses latest version of golangci-lint and we wanted to avoid that to prevent broken GHAs when a new release is incompatible with the repo. We left it in workflow file so that dependabot can manage its updates.
https://github.com/linode/linode-cloud-controller-manager/pull/226#issuecomment-2538367324

Andrews2024: In this particular instance it doesn't seem to be an issue with versioning, but rather some discrepancy between the CLI tool and the GHA. Using v1.64.5 locally doesn't produce the same error as the GHA.

rahulait (approved):

PR #352, opened 2025-03-11
Disable service account when rbac disabled

rammanoj:
### General:

Disable service account when rbac disabled

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/deploy/chart/templates/daemonset.yaml
+++ b/deploy/chart/templates/daemonset.yaml
@@ -18,7 +18,9 @@ spec:
       labels:
         app: ccm-linode
     spec:
+      {{- if ne .Values.rbacEnabled false }}
       serviceAccountName: ccm-linode
+      {{- end }}
       {{- with .Values.securityContext }}
       securityContext:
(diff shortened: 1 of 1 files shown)


deploy/chart/templates/daemonset.yaml
```diff
@@ -35,7 +37,11 @@ spec:
       tolerations:
       {{- toYaml . | nindent 8 }}
       {{- end }}
-      hostNetwork: {{ default true .Values.enableHostNetwork }}
+      {{- if ne .Values.enableHostNetwork false }}
+      hostNetwork: true
+      {{- else }}
+      hostNetwork: false
+      {{- end }}
```
rahulait, line 44: Do we need this change? Can we just make sure we have false and not "false" in values.yaml to make sure its set correctly?

rammanoj: fair enough. I was trying to get the value commented in values.yaml. But I think having a definition is much more simpler.

rahulait (approved):

PR #357, opened 2025-03-17
[test] Subnet filter testing

Andrews2024:
### General:
This PR adds a chainsaw test to verify that CCM correctly manages nodes on a subnet specified by the `--subnet-names` flag. There is a new make target for setting up the second cluster on a second subnet of the VPC used by the first cluster. The first cluster should manage nodes on the default subnet, while the second cluster should manage nodes on the testing subnet.

### Pull Request Guidelines:

1. [X] Have you added tests? 


Makefile
```diff
@@ -14,16 +14,21 @@ HELM_VERSION            ?= v3.16.3
 # Dev Setup
 #####################################################################
 CLUSTER_NAME            ?= ccm-$(shell git rev-parse --short HEAD)
+SUBNET_CLUSTER_NAME     ?= subnet-testing
```
rahulait, line 17: this might have issues if there are two PRs running this test at the same time. We'll need unique cluster names.

Makefile
```diff
@@ -207,6 +213,27 @@ e2e-test-bgp:
...
+e2e-test-subnet: 
+	# Generate cluster manifests for second cluster
+	SUBNET_NAME=testing CLUSTER_NAME=$(SUBNET_CLUSTER_NAME) MANIFEST_NAME=$(SUBNET_MANIFEST_NAME) VPC_NAME=$(CLUSTER_NAME) \
+		VPC_NETWORK_CIDR=172.16.0.0/16 K8S_CLUSTER_CIDR=172.16.64.0/18 make generate-capl-cluster-manifests
+	# Follow create-capl-cluster steps minus NodeHealthy=true step (times out otherwise)
+	kubectl apply -f $(SUBNET_MANIFEST_NAME).yaml
+	kubectl wait --for=condition=ControlPlaneReady cluster/$(SUBNET_CLUSTER_NAME) --timeout=600s || (kubectl get cluster -o yaml; kubectl get linodecluster -o yaml; kubectl get linodemachines -o yaml)
+	clusterctl get kubeconfig $(SUBNET_CLUSTER_NAME) > $(SUBNET_KUBECONFIG_PATH)
+	KUBECONFIG=$(SUBNET_KUBECONFIG_PATH) kubectl wait --for=condition=Ready nodes --all --timeout=600s
+	KUBECONFIG=$(SUBNET_KUBECONFIG_PATH) kubectl taint nodes -l node-role.kubernetes.io/control-plane node-role.kubernetes.io/control-plane-
```
rahulait, line 226: Can we re-use `make create-capl-cluster` by setting correct env vars here?

Andrews2024: I found this step of `create-caple-cluster` consistently timed out when setting up the second cluster: `kubectl wait --for=condition=NodeHealthy=true machines -l cluster.x-k8s.io/cluster-name=$(CLUSTER_NAME) --timeout=900s`. It seemed to fail even when the cluster itself was actually usable for testing.

Makefile
rahulait: yeah. We need to make sure we don't exceed the name limits. Example: https://github.com/linode/cluster-api-provider-linode/blob/main/e2e/capl-cluster-flavors/kubeadm-capl-cluster/chainsaw-test.yaml#L18-L21

Makefile
rahulait: Yeah that will happen because CCM is not started with `--subnet-names` flag set correctly, so it will fail to set node addresses as they won't be there in cache. We'll need to patch the generated yaml before apply to set CCM's subnet name in it and then it should work fine.

Makefile
```diff
@@ -207,6 +213,29 @@ e2e-test-bgp:
...
+	# Add subnetNames to HelmChartProxy	
+	yq e 'select(.kind == "HelmChartProxy" and .spec.chartName == "ccm-linode").spec.valuesTemplate' $(SUBNET_MANIFEST_NAME).yaml > tmp.yaml
+	yq -i e '.routeController  += {"subnetNames": "testing"}' tmp.yaml
+	yq -i e '.routeController.vpcNames = "{{.InfraCluster.spec.vpcRef.name}}"' tmp.yaml
+	yq -i e 'select(.kind == "HelmChartProxy" and .spec.chartName == "ccm-linode").spec.valuesTemplate = load_str("tmp.yaml")' $(SUBNET_MANIFEST_NAME).yaml
+	rm tmp.yaml
+	# Create the second cluster
+	MANIFEST_NAME=$(SUBNET_MANIFEST_NAME) CLUSTER_NAME=$(SUBNET_CLUSTER_NAME) KUBECONFIG_PATH=$(SUBNET_KUBECONFIG_PATH) \
+		make create-capl-cluster
+	# Patch both cluster CCM daemonsets with --subnet-names
```
rahulait, line 230: I think you need this as well?
```
KUBECONFIG_PATH=$(SUBNET_KUBECONFIG_PATH) make patch-linode-ccm
```

rahulait:
Since we use pull_request_target, checks run against the PR branch don't run the subnet test. Once we merge the PR, this will start running as well. LGTM

rahulait (approved):

PR #366, opened 2025-03-31
[feat] : enable nodeipamcontroller within CCM

rahulait:
### General:
This PR enables CCM to also do node IPAM management. To enable nodeipam-controller, one needs to set `--allocate-node-cidrs` along with `cluster-cidr` flag.

```yaml
spec:
  template:
    spec:
      containers:
        - name: ccm-linode
          args:
            - --allocate-node-cidrs=true
            - --cluster-cidr=10.192.0.0/10
```

### Pull Request Guidelines:

1. [x] Have you added tests? 


cloud/linode/nodeipamcontroller.go
```diff
@@ -0,0 +1,167 @@
...
+	ctx := wait.ContextForChannel(stopCh)
+
+	nodeIpamController, err := nodeipamcontroller.NewNodeIpamController(
+		ctx,
+		nodeInformer,
+		cloud,
+		kubeclient,
+		clusterCIDRs,
+		serviceCIDR,
+		secondaryServiceCIDR,
```
komer3, line 112: question: Do we need to pass this empty pointer? We aren't setting it yet since the code block above this commented

rahulait: This is modified version of upstream https://github.com/kubernetes/kubernetes/blob/master/cmd/cloud-controller-manager/nodeipamcontroller.go. Even there, if the flag is not set, empty pointer is passed and underlying code does the handling. So we should be good as we don't intend to set or use it.

komer3 (approved):
LGTM!

PR #423, opened 2025-07-22
[improvement] : allow passing vpcids and subnetids to ccm

rahulait:
### General:

This PR adds two new flags to CCM (`--vpc-ids` and `--subnet-ids`). This will allow someone to specify ids instead of names when starting CCM.

Key changes:
Added --vpc-ids and --subnet-ids command-line flags as IntSlice types
Changed existing --vpc-names and --subnet-names flags from string to StringSlice types
Added validation logic to ensure only one approach (names or IDs) is used at a time

### Pull Request Guidelines:

1. [x] Have you added tests? 

### DIFF
--- a/docs/configuration/environment.md
+++ b/docs/configuration/environment.md
@@ -34,28 +34,28 @@ The CCM can be configured using environment variables and flags. Environment var
 
 The CCM supports the following flags:
 
-| Flag | Default | Description |
-|------|---------|-------------|
-| `--linodego-debug` | `false` | Enables debug output for the LinodeAPI wrapper |
-| `--enable-route-controller` | `false` | Enables route_controller for CCM |
-| `--enable-token-health-checker` | `false` | Enables Linode API token health checker |
-| `--vpc-names` | `[]` | Comma separated VPC names whose routes will be managed by route-controller |
-| `--subnet-names` | `["default"]` | Comma separated subnet names whose routes will be managed by route-controller (requires vpc-names flag) |
-| `--vpc-ids` | `[]` | Comma separated VPC ids whose routes will be managed by route-controller |
-| `--subnet-ids` | `[]` | Comma separated subnet ids whose routes will be managed by route-controller (requires vpc-ids flag) |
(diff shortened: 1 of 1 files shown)


cloud/linode/vpc.go
```diff
@@ -170,3 +167,107 @@ func getNodeBalancerBackendIPv4SubnetID(client client.Client) (int, error) {
...
+	subnetNames := []string{}
+	for _, subnetID := range Options.SubnetIDs {
+		subnet, err := client.GetVPCSubnet(context.TODO(), vpcID, subnetID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get subnet %d for VPC %d: %w", subnetID, vpcID, err)
+		}
+
+		subnetIDs[subnet.Label] = subnet.ID
+		subnetNames = append(subnetNames, subnet.Label)
+	}
```
scmeggitt, line 226: won't this still lead to the same problem if the subnet name is changed? Since we are just grabbing the label at one point in time?

rahulait: Nope. We store everything in the cache and then it always looks into the cache.
vpcIDs and subnetIDs are the two caches it builds at startup and then uses going forward.

rahulait: Nope. If someone updates vpcname/subnetname and CCM is restarted, it will again try to get the label for the passed vpcid/subnetid and store the updated name in its cache and continue working. When logging, it will log the vpc/subnet names instead of ids which are more readable/human friendly than the ids.

scmeggitt: ahh ok I missed the vpcID logic since that was already there. Makes a lot more sense to me now.

Only other comment would be why can't you use ids and names?

rahulait: You mean flipping the logic and using ids instead of names? Or you meant combination of vpcname+subnetid?
I didn't flip the logic as that would mean all existing tests would start failing and that would be a much bigger change and touch most of the functionality of CCM.

scmeggitt: the combination, but it really isn't that important. One or the other is fine with me.

komer3 (approved):

PR #437, opened 2025-08-14
[feat] Add IP Reservation support for Services backed by NodeBalancers

hcwagner:
This PR is focused on adding support for IP Reservation while using NodeBalancers as a Service LoadBalancer. This assume that the customer has:

Reserved an IP through Linode CLI, CM, terraform, etc.
Supplies the reserved IP attached to their account through the provided Annotation


komer3 (approved):
LGTM!

PR #561, opened 2026-04-17
[test] Stabilize flaky e2e tests by replacing hard sleeps with polling loops

moshevayner:
## What

- Fix `get-nb-id.sh` to read the NodeBalancer IP directly from `status.loadBalancer.ingress[0].ip` instead of parsing it from the service hostname DNS label, which was brittle and caused widespread test failures.

- Replace hard `sleep` + one-shot checks with polling loops (`for i in {1..N}; do ... sleep 10; done`) across 15+ tests. Hard sleeps either fail when the system is slow or waste time when it's fast.

- Fix the old hostname-based NB ID lookup in
  `lb-with-proxyprotocol-default-annotation` (same root cause as `get-nb-id.sh`).

- Fix `name: c` typo in `lb-created-with-reserved-ip-multiple-change-ip`.

- Remove `sleep 180` in `lb-with-node-addition` before the node provisioning poll; extend the loop from 20→30 iterations to keep the same ~10m budget without a blind wait.

## Why

These tests were flaky on timing, not because of real CCM bugs. The polling pattern is already established in the codebase. This PR applies it consistently.

Signed-off-by: Moshe Vayner <moshe@vayner.me>

### General:

* [ ] Have you removed all sensitive information, including but not limited to access keys and passwords?

### Pull Request Guidelines:

1. [ ] Have you added tests? 

### DIFF
--- a/e2e/test/lb-with-node-addition/chainsaw-test.yaml
+++ b/e2e/test/lb-with-node-addition/chainsaw-test.yaml
@@ -82,6 +82,8 @@ spec:
               required_replicas=$((current_replicas + 1))
               KUBECONFIG=$MGMT_KUBECONFIG kubectl patch machinedeployment ${CLUSTER_NAME}-md-0 --type='merge' -p "{\"spec\":{\"replicas\":$required_replicas}}"
 
+              sleep 180
+
               nbid=$(KUBECONFIG=$KUBECONFIG NAMESPACE=$NAMESPACE LINODE_TOKEN=$LINODE_TOKEN ../scripts/get-nb-id.sh)
               echo "Nodebalancer ID: $nbid"
 
--- a/e2e/test/lb-with-node-addition/chainsaw-test.yaml
+++ b/e2e/test/lb-with-node-addition/chainsaw-test.yaml
@@ -82,12 +82,10 @@ spec:
               required_replicas=$((current_replicas + 1))
               KUBECONFIG=$MGMT_KUBECONFIG kubectl patch machinedeployment ${CLUSTER_NAME}-md-0 --type='merge' -p "{\"spec\":{\"replicas\":$required_replicas}}"
 
-              sleep 180
-
               nbid=$(KUBECONFIG=$KUBECONFIG NAMESPACE=$NAMESPACE LINODE_TOKEN=$LINODE_TOKEN ../scripts/get-nb-id.sh)
               echo "Nodebalancer ID: $nbid"
 
-              for i in {1..20}; do
+              for i in {1..30}; do
                   response=$(curl -sf \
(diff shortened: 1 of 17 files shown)


e2e/test/lb-with-node-addition/chainsaw-test.yaml
```diff
@@ -82,12 +82,10 @@ spec:
               required_replicas=$((current_replicas + 1))
               KUBECONFIG=$MGMT_KUBECONFIG kubectl patch machinedeployment ${CLUSTER_NAME}-md-0 --type='merge' -p "{\"spec\":{\"replicas\":$required_replicas}}"
 
-              sleep 180
-
               nbid=$(KUBECONFIG=$KUBECONFIG NAMESPACE=$NAMESPACE LINODE_TOKEN=$LINODE_TOKEN ../scripts/get-nb-id.sh)
               echo "Nodebalancer ID: $nbid"
 
-              for i in {1..20}; do
+              for i in {1..30}; do
```
komer3, line 90: We should keep the large sleep since we are using CAPL to create a new node, and that can take ~2 mins to come up ready. Otherwise, using polling through out the tests cases is great!

moshevayner: So, I basically kept it at the same length by increasing the counter to 30 from 20. Adding 10 more iterations of 20 seconds each is 200 seconds. I just made it a bit more flexible so that if the CAPL node comes up in less than 180 seconds, we save the time and move on to the next step as soon as everything's ready.
The math I did was that instead of waiting 180 seconds for CAPI, and then looping for 400 seconds max (total of 580 seconds), we just start the loop right away, but with a higher max of 600 seconds total, which even gives us an extra 20 seconds if we need them.
Does that make sense? If you don't think that's the right approach, I can definitely revert this change.
It's just that this test was sometimes failing so I figured I'd give it a shot with being a bit more flexible with the loop.
WDYY?

komer3: Oh we definitely still wanna keep the polling. Just start it later. 

Reason being that hammering the API when we know the state won't change might not be what we wanna do? We know that a node can take a minimum of 2 mins to come up, so why not sleep for that time and then start polling to reduce API calls?

We are also running other tests in parallel. Potentially running multiple GHA runs in parallel. I think it's better to reduce API calls whenever we can. Reduces our chances of hitting rate limits, timeouts, etc.

moshevayner: Yeah, that makes total sense! I'll revert this one. 
Thanks for catching that!

komer3 (approved):

PR #620, opened 2026-08-31
[fix]: Chunked cloud firewall rules of only 1 addr type result in empty rule and failed updates

wbh1:
### General:

### Pull Request Guidelines:

1. [x] Have you added tests? 

### Description

tl;dr — if you have a large firewall that exceeds the 255 address limit and kicks in the chunking logic, if that large firewall rule contains only 1 address type (i.e. all IPv4 or all IPv6) then the CCM still tries to add an empty rule for the address type.

This does the dumb, simple thing of just checking the length of the rules for each addr type before chunking them.

Example with a big rule of only IPv4 addrs:

```console
 I0831 16:54:33.298782   37901 firewalls.go:305] Firewall label 'ACCEPT-platform-components-gateway-istio' is too long. Stripping to
 'ACCEPT-platform-components-gatew'
 inbound rules: 3
 rule 0: IPv4=255 IPv6=0
 rule 1: IPv4=9 IPv6=0
 rule 2: IPv4=0 IPv6=0
```

### DIFF
--- a/cloud/linode/services/firewalls.go
+++ b/cloud/linode/services/firewalls.go
@@ -265,8 +265,11 @@ func chunkIPs(ips []string) [][]string {
 	ipCount := len(ips)
 
 	// If the number of IPs is less than or equal to maxIPsPerFirewall,
-	// return a single chunk containing all IPs.
+	// return a single chunk containing all IPs, or nil if chunking an empty slice.
 	if ipCount <= maxIPsPerFirewall {
+		if ipCount == 0 {
+			return nil
+		}
 		return [][]string{ips}
 	}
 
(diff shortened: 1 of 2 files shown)


AshleyDumaine:
> This does the dumb, simple thing of just checking the length of the rules for each addr type before chunking them.

While this change works, I'm thinking we'll want to fix chunkIPs to actually return nil if the length of the ips is zero instead of just adding the checks here. WDYT?

wbh1:
> While this change works, I'm thinking we'll want to fix chunkIPs to actually return nil if the length of the ips is zero instead of just adding the checks here. WDYT?

Good point — I switched the approach so that chunkIPs will return `nil` when called using an empty slice as the arg to the function.

AshleyDumaine (approved):

PR #632, opened 2026-09-22
[feat] add support for reserved IPv4 range in NodeBalancer backend allocation

vsilva-akamai:
### Summary

- Adds `--nodebalancer-backend-ipv4-reserved-range` and the matching Helm value for reserving the highest `/30` in a configured NodeBalancer backend subnet.
- Allocates the first available `/30` below the reserved range by reading the subnet's existing NodeBalancer assignments.
- Rejects invalid reservations, existing assignments that use the reserved range, and service-provided backend ranges that overlap the reservation.
- Adds unit coverage for allocation, validation, VPC option construction, and Helm rendering.

### Rationale

Reserving the highest `/30` prevents CCM-managed NodeBalancer backend allocation from consuming an address range that must remain available for networking.

## Validation

- `go test ./cloud/linode -count=1`
- `make helm-template`

### General

### Pull Request Guidelines

1. [x] Have you added tests?


deploy/chart/templates/daemonset.yaml
```diff
@@ -135,10 +135,16 @@ spec:
...
+            {{- if hasKey .Values.routeController "configureCloudRoutes" }}
+            - --configure-cloud-routes={{ .Values.routeController.configureCloudRoutes }}
+            {{- else }}
+            - --configure-cloud-routes=true
+            {{- end }}
             {{- with .Values.routeController.routeReconciliationPeriod }}
             - --route-reconciliation-period={{ . }}
             {{- end }}
+            {{- else if hasKey .Values "configureCloudRoutes" }}
+            - --configure-cloud-routes={{ .Values.configureCloudRoutes }}
```
AshleyDumaine: Doesn't this make it so `--configure-cloud-routes` is passed twice if values has `routeController.configureCloudRoutes` and `configureCloudRoutes` set?
If we want to add support to configure cloud routes outside of enabling the route controller, we might want to deprecate/remove the option from the route controller itself.

AshleyDumaine:
Can you document the new feature on `docs/configuration/environment.md` and `docs/configuration/loadbalancer.md` as well?

deploy/chart/templates/daemonset.yaml
vsilva-akamai: I teested this with both keys set to opposite values to check which one would be  the winner, only one `--configure-cloud-routes` renders, and it's the routeController one. 
L146 is the else if arm of the {{- if .Values.routeController }} at L124, so the two branches don't both fire. Letting the top-level false pass would render `--enable-route-controller=true` `--configure-cloud-routes=false`, which stops the route controller from start.

vsilva-akamai:
```
$ helm template ccm ./deploy/chart \
    --set configureCloudRoutes=false \
    --set routeController.configureCloudRoutes=true \
    --set routeController.vpcNames=vpc1 --set routeController.clusterCIDR=10.0.0.0/8
```

```
          name: ccm-linode
          args:
            - --leader-elect-resource-lock=leases
            - --v=3
            - --secure-port=10253
            - --webhook-secure-port=0
            - --enable-route-controller=true
            - --configure-cloud-routes=true
            - --vpc-names=vpc1
            - --cluster-cidr=10.0.0.0/8
```

vsilva-akamai:
> Can you document the new feature on `docs/configuration/environment.md` and `docs/configuration/loadbalancer.md` as well? I think we'll also want a new chainsaw e2e test for this.

I updated the docs as requested, and now I'm working on getting the Chainsaw tests implemented and working.

